PR #28, opened 2023-08-07
add workflow to test build

struga0258:


.github/workflows/build_test.yml
```diff
@@ -0,0 +1,52 @@
+name: Build and Test Workflow
```
1000TurquoisePogs: No "test" seen

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,52 @@
...
+on:
+  push:
+    branches:
```
1000TurquoisePogs, line 4: how about on manual trigger too, or not?

1000TurquoisePogs (commented):
Glad this is working for you but note these comments, and then consider this: https://github.com/TSTerminal/TSTerm/tree/main/.github/workflows
Why not make a 2 workflow approach as seen here: 1) on every PR, build the assets and upload to the github workflow
2) on merge to main, do a "draft release"

it means you automatically get nice content like this https://github.com/TSTerminal/TSTerm/releases

1000TurquoisePogs (approved):

PR #29, opened 2023-08-17
Schema to Form

sakshibobade21:
This PR is about adding the HTML forms which are dynamically derived from the schema.


src/renderer/components/common/JsonForms.tsx
```diff
@@ -10,23 +10,160 @@
...
+const makeUISchema = (schema: any, base: string, formData: any): any => {
   const properties = Object.keys(schema.properties);
-  const elements = properties.map((p: any) => {
-    if (schema.properties[p].type === 'object') {                                 
-      return makeUISchema(schema.properties[p], `${base}${p}/properties/`);
+
+  const elements = properties.map((prop: any) => {
+    if (schema.properties[prop].type === 'object') {
+
+      if(formData && formData.type && ((formData.type != 'PKCS12' && prop == 'pkcs12') || (formData.type == 'PKCS12' && prop == 'keyring'))) {
```
1000TurquoisePogs: this is a specific case present in a reusable component.
this cannot be the correct way to solve this problem.
this problem will come up again in other parts of config, so there needs to be a reusable solution.

1000TurquoisePogs (changes requested):

1000TurquoisePogs (approved):
This looks good.

sakshibobade21:
Closing this PR. Here is the new one addressing the same concern but with additional features

https://github.com/zowe/zen/pull/43

PR #30, opened 2023-08-21
UI to Upload Pax for Offline Zowe Install

timgerstel:

- Use native system dialog to allow users to select PAX files
- Store path as state

Todo: do something with the selected file


1000TurquoisePogs (commented):
Concept is fine.
please target v2.x/staging
not sure if the install button is in the right place at this time / finished, but code you have so far makes sense.

timgerstel:
Work was included in https://github.com/zowe/zen/pull/31/files

PR #31, opened 2023-08-25
Install Type Selection Stage

timgerstel:

- Add a stage which allows the user to select from the following installation types:
  - Download Zowe PAX from internet (not yet implemented)
  - Upload offline PAX (needs FTP upload functionality)
  - SMP/E install (shows message that SMP/E install is unsupported via Zen and must be completed manually)

### DIFF
--- a/src/renderer/components/stages/installation/InstallTypeSelection.tsx
+++ b/src/renderer/components/stages/installation/InstallTypeSelection.tsx
@@ -17,7 +17,8 @@ import { selectInstallationArgs, selectZoweVersion, setInstallationArgs } from '
 import { selectConnectionArgs } from '../connection/connectionSlice';
 import JsonForm from '../../common/JsonForms';
 import { IResponse } from '../../../../types/interfaces';
-import ProgressCard from '../../common/ProgressCard'
+import ProgressCard from '../../common/ProgressCard';
+import CheckCircleOutlineIcon from '@mui/icons-material/CheckCircleOutline';
 
 const InstallationType = () => {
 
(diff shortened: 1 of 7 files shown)


1000TurquoisePogs:
This looks unfinished to me so I'm not yet merging it.
There IS a download action already in ZEN, so if upload or SMPE are selected, then the download should not occur. But I think it still does, because I didn't see any code that prevents that.

SMPE says "unsupported" but doesnt really need to be - if the user is prompted for the install location, then SMPE is "supported" in that ZEN can work in its presence.

I'd expect SMPE to do:
* unhide an input box requiring stating the location of where Zowe was installed by SMPE
* Print the message "SMP/E installation must be done outside of ZEN. Return to ZEN and input the location Zowe was installed to before continuing"

src/renderer/components/stages/installation/InstallTypeSelection.tsx
```diff
@@ -0,0 +1,111 @@
...
+        >
+            <FormControlLabel value="download" control={<Radio />} label="Download Zowe convenience build PAX from internet" />
+            <FormControlLabel value="upload" control={<Radio />} label="Upload Zowe PAX for offline install" />
+            <FormControlLabel value="smpe" control={<Radio />} label="SMP/E" />
+        </RadioGroup>
+    </FormControl>
+    {installValue === "smpe" && <Typography id="position-2" sx={{ mb: 1, whiteSpace: 'pre-wrap' }} color="text.secondary">       
+        {`SMP/E installation must be done outside of ZEN. Return to ZEN and input the location Zowe was installed to before continuing.`}
+    </Typography>}
+    {installValue === "smpe" && <FormControl>
```
timgerstel, line 77: add button to validate smpe location

DivergentEuropeans (approved):
Tim and I reviewed this together and changes have been addressed

PR #33, opened 2023-08-28
Add the getters and setters for the zowe.yaml

sakshibobade21:
This PR is created to add the config service for getter/setters for the subsections of the zowe.yaml.
And integrating this service with the HTML forms dynamically derived from the Schema.


1000TurquoisePogs (approved):
Looks good to me.
About flat being added: this is used elsewhere in Zowe already, so it is OK.

DivergentEuropeans (changes requested):
Moving to Page 2 - "Planning" section, I get run-time error

image

sakshibobade21:
Here is the new PR, addressing the same concerns but with enhancements:

https://github.com/zowe/zen/pull/43

PR #35, opened 2023-08-29
Release workflow zen

Himani1519:
Github action for release workflow of zen.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release zen draft
 on:
   workflow_run:
     workflows: ["Build zen"]
-    branches: [main]
+    branches: [v2.x/staging]
     types:
       - completed
   workflow_dispatch:
@@ -38,8 +38,8 @@ jobs:
         if: ${{ inputs.RUN_ID == '' }}
         env:
           GH_TOKEN: ${{ github.token }}
-        # download default by filtering for the latest successful build output by the main branch
-        run: gh run download $(gh run list -w "Build zen" --json databaseId,headBranch,conclusion -q 'map(select(.headBranch == "main")) | map(select(.conclusion == "success")) | .[0].databaseId')
+        # download default by filtering for the latest successful build output by the v2.x/staging branch
+        run: gh run download $(gh run list -w "Build zen" --json databaseId,headBranch,conclusion -q 'map(select(.headBranch == "v2.x/staging")) | map(select(.conclusion == "success")) | .[0].databaseId')
       - name: '[Download] Input'
         if: ${{ inputs.RUN_ID }}
         env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release zen draft
 on:
   workflow_run:
     workflows: ["Build zen"]
-    branches: [v2.x/staging]
+    branches: [v2.x/main]
     types:
       - completed
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,66 @@
+name: Release zen draft
+on:
+  workflow_run:
+    workflows: ["Build zen"]
+    branches: [main]
```
struga0258: we dont have main, we have v2.x/main

.github/workflows/release.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          echo "BUMP_TYPE=${{ inputs.BUMP_TYPE }}" >> $GITHUB_ENV
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: '[Download] Default'
+        if: ${{ inputs.RUN_ID == '' }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        # download default by filtering for the latest successful build output by the main branch
+        run: gh run download $(gh run list -w "Build zen" --json databaseId,headBranch,conclusion -q 'map(select(.headBranch == "main")) | map(select(.conclusion == "success")) | .[0].databaseId')
```
struga0258: Also here

.github/workflows/release.yml
```diff
@@ -0,0 +1,66 @@
...
+          GH_TOKEN: ${{ github.token }}
+        # download default by filtering for the latest successful build output by the  v2.x/main branch
+        run: gh run download $(gh run list -w "Build zen" --json databaseId,headBranch,conclusion -q 'map(select(.headBranch == "v2.x/main")) | map(select(.conclusion == "success")) | .[0].databaseId')
+      - name: '[Download] Input'
+        if: ${{ inputs.RUN_ID }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh run download ${{ inputs.RUN_ID }}
+      - name: '[Archive] zip download'
+        run: zip -r zen.zip zen
```
1000TurquoisePogs: a zip is here, but what does it contain? i dont think it contains anything actually, because i dont think the `gh run download` has anything to download, because `build zen` doesnt upload anything.

.github/workflows/release.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: '[Create release]'
+        uses: softprops/action-gh-release@v1
+        with:
+          tag_name: ${{ steps.tag_version.outputs.new_tag }}
+          name: Release ${{ steps.tag_version.outputs.new_tag }}
+          draft: true
+          prerelease: false
+          token: ${{ github.token }}
+          #body_path: CHANGELOG.md
+          files: zen.zip
```
1000TurquoisePogs: this would be wrong due to there being no zen.zip

1000TurquoisePogs (changes requested):
this doesnt appear to work because the `build zen` step is not uploading builds to github, it is uploading them to artifactory.
this script is trying to download from a previous github upload, instead of an artifactory upload.

question is, should the release even have an upload anyway?
i don't think its necessary. skip the zip download & file upload part. the release otherwise publishes the source code i believe, which is enough.

Himani1519:
> this doesnt appear to work because the `build zen` step is not uploading builds to github, it is uploading them to artifactory. this script is trying to download from a previous github upload, instead of an artifactory upload.
> 
> question is, should the release even have an upload anyway? i don't think its necessary. skip the zip download & file upload part. the release otherwise publishes the source code i believe, which is enough.

Make sense, I will update the actions

struga0258 (approved):

1000TurquoisePogs (approved):

PR #42, opened 2023-09-12
Integrate Monaco Editor in ZEN

sakshibobade21:
1) Incorporate Monaco Editor to ZEN with the Schema and Yaml Validations for the entire zowe.yaml, not just the subschema
2) Integrating the editor’s with the configuration service
3) Automatically update the JSON forms with the corresponding editor’s content if it adheres to schema and yaml validation
4) Enable bidirectional synchronization: changes made in the form must reflect in the editor
5) Centralized the shared code for the editor


src/services/configService.ts
```diff
@@ -1,13 +1,32 @@
 import { flatten, unflatten } from 'flat';
```
timgerstel: Why is there two configservice.ts? One is ConfigService.ts the other is configService.ts

timgerstel: Looks like this file should just be deleted

timgerstel:
Going through testing this PR, the new UI for the installation config stage looks sooo much better!
image

It does look like its missing some details though. It looks like maybe the first section is missing a title, and "parmlibMembers" isnt a great section title either. This stage also used to have details filled out on the first from the example-zowe.yaml which no longer occurs as you can see in the screenshot above. I would consider this a regression and ask that it be fixed!

timgerstel:
Tested the schema validation and it works as expected, rejects things that do not adhere:

image


On this UI my one note is that the editor should be autofocused when this popup appears, so if you press Ctrl+A on your keyboard all the text highlights without having to click the editor first.

timgerstel:
Overall i think this is a great addition, just a few minor things before I approve

src/services/configService.ts
timgerstel: This double file messed up my git shell so bad lol i cant switch branches now

timgerstel:
The installation stage doesnt autofill details from the example-zowe.yaml on the first run anymore btw, this may be a regression

timgerstel:
There is another bug with this PR. I wanted to retest the autofill for the first run, so I deleted all of my cached data in the userData folder for Zen (%appdata/zowe-enterprise-necessity on Windows I think) and now I get this error:
image

This does not happen on v2.x/staging. It looks like your code might not be handling the case where a yaml doesnt exist at all

sakshibobade21:
Created a new one.

PR #47, opened 2023-09-25
Add Text Modification Operations to Dev Mode + Production Mode Context Menus

timgerstel:

- Introduce cut, copy, paste, undo, redo, and select all to Zen's context menu for both dev mode and production mode


src/main/menu.ts
```diff
@@ -44,6 +44,32 @@ export default class MenuBuilder {
...
+      { role: 'redo' },
+      { type: 'separator' },
+      { role: 'cut' },
+      { role: 'copy' },
+      { role: 'paste' },
+      { type: 'separator' },
+      { role: 'selectAll' }
+    ])
+
+    this.mainWindow.webContents.on('context-menu', (_, props) => {
```
DivergentEuropeans, line 58: You're adding event listeners to `this.mainWindow.webContents`, but you're not removing them. I'm not very familiar with Electron's lifecycle, so it may do this already, but to be safe, it's better to add later on in the code:
`this.mainWindow.webContents.removeListener('context-menu', handleContextMenu);`

src/main/menu.ts
```diff
@@ -44,6 +44,32 @@ export default class MenuBuilder {
     const menu = Menu.buildFromTemplate(template);
     Menu.setApplicationMenu(menu);
 
+    const selectAllAndCopyMenu = Menu.buildFromTemplate([
```
DivergentEuropeans: `selectAllAndCopyMenu` isn't being used is it? Did you intend to leave it - maybe a comment and a TODO for a later time?

DivergentEuropeans (changes requested):
A pair of small things

src/main/menu.ts
timgerstel: I think it gets cleaned up when the app closes, there is no context when the app is running where you wouldnt want this event listener

src/main/menu.ts
timgerstel: I had this in logic that i removed, ill remove the unused code good catch

src/main/menu.ts
timgerstel: The electron boilerplate code for the dev environment doesnt even remove it

DivergentEuropeans (approved):
Nice, looks good

DivergentEuropeans:
Code makes sense now. I'll test in a day or two & we can merge

1000TurquoisePogs:
works on mac but not linux, filing bug.

PR #51, opened 2023-10-02
Set correct run-time directory from SMPE install type

timgerstel:

- Properly track smpe directory via user input
- Prevent installation stage from creating datasets/downloading pax/running 'zwe init' if the user has chosen SMPE install type (these steps will have been done by the SMPE process)
- Update wording in description of SMPE dir input field


1000TurquoisePogs (commented):
I cannot properly test this due to an error I am getting

Error occurred in handler for 'set-config-by-key': TypeError: Cannot read properties of null (reading 'properties')
    at ConfigurationStore.validateWithSchema (/home/me/bt/zen/.webpack/main/index.js:30877:33)
    at ConfigurationStore.setConfigByKey (/home/me/bt/zen/.webpack/main/index.js:30905:18)
    at PlanningActions.<anonymous> (/home/me/bt/zen/.webpack/main/index.js:29866:68)
    at Generator.next (<anonymous>)
    at /home/me/bt/zen/.webpack/main/index.js:29745:71
    at new Promise (<anonymous>)
    at __webpack_modules__../src/actions/PlanningActions.ts.__awaiter (/home/me/bt/zen/.webpack/main/index.js:29741:12)
    at PlanningActions.setConfigByKey (/home/me/bt/zen/.webpack/main/index.js:29865:16)
    at /home/me/bt/zen/.webpack/main/index.js:29971:61
    at Generator.next (<anonymous>)
[2898772:1003/131914.703293:INFO:CONSOLE(0)] "Uncaught (in promise) Error: Error invoking remote method 'set-config-by-key': TypeError: Cannot read properties of null (reading 'properties')", source: http://localhost:3000/main_window#/wizard (0)


But I now believe the SMPE runtime directory prompt to be redundant, so would like to test.


The radio to select SMPE/download/upload options comes AFTER you specify where you will install zowe to, which is the previous page.
Due to this, I think you're actually prompting the user twice for the same info.

Further, with SMPE we also need to know where datasets were put but again I think that's asked on the next page, anyway.

image

image


image

sakshibobade21:
The PR looks good to me. Please resolve the conflicts before merging.

sakshibobade21 (approved):

DivergentEuropeans (approved):
There's a small issue in that Validate Location when validation fails, does not give an error or some UX indication

image


But not a problem with the PR @timgerstel just a feature request so I've updated this PR w/ merge conflicts & going to merge

PR #65, opened 2023-10-24
Updating planning stage

sakshibobade21:
Updating the planning stage and adding the additional fields to it


1000TurquoisePogs (changes requested):
Looks like a good UI, but it seems to be missing the default values.
I would not expect users to want to customize the ID fields normally, so they do need default values. Same for Job fields.

For "Installation location" header, can we rename to "Installation location (Runtime directory)"? or is that too long?
Because thats what we call it, and what our documentation calls it.

1000TurquoisePogs (approved):

PR #45, opened 2023-09-18
Added UI for status indicators + change to use Node-native https call

DivergentEuropeans:
image

other styles include:

image

Why change to Node's https? Because fetch is not available on Windows in Electron apps

PR #83, opened 2023-11-09
Enhance Installation Stage to Run 'zwe init mvs'

timgerstel:
## Proposed changes

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/src/actions/InstallationHandler.ts
+++ b/src/actions/InstallationHandler.ts
@@ -62,11 +62,21 @@ class Installation {
       const unpax = await this.unpax(connectionArgs, installationArgs.installationDir); 
       ProgressStore.set('installation.unpax', unpax.status);
 
-      console.log("installing...");
-      const install = await this.install(connectionArgs, installationArgs.installationDir);
-      ProgressStore.set('installation.install', install.status);
+      let installation;
+      if(installationArgs.installationType !== "smpe"){
+        console.log("installing...");
+        const install = await this.install(connectionArgs, installationArgs.installationDir);
+        ProgressStore.set('installation.install', install.status);
+      } else {
+        installation = {status: true}
+        ProgressStore.set('installation.install', true);
+      }
 
-      return {status: download.status && uploadYaml.status && upload.status && unpax.status && install.status, details: ''};
+      console.log("running zwe init mvs...");
+      const initMvs = await this.initMVS(connectionArgs, installationArgs.installationDir);
+      ProgressStore.set('installation.initMVS', initMvs.status);
+
+      return {status: download.status && uploadYaml.status && upload.status && unpax.status && installation.status && initMvs.status, details: ''};
     } catch (error) {
       return {status: false, details: error.message};
     }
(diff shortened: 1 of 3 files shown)


src/actions/InstallationHandler.ts
```diff
@@ -62,11 +62,21 @@ class Installation {
...
+        console.log("installing...");
+        const install = await this.install(connectionArgs, installationArgs.installationDir);
+        ProgressStore.set('installation.install', install.status);
+      } else {
+        installation = {status: true}
+        ProgressStore.set('installation.install', true);
+      }
+
+      console.log("running zwe init mvs...");
+      const initMvs = await this.initMVS(connectionArgs, installationArgs.installationDir);
```
1000TurquoisePogs: it looks to me like init mvs is run regardless of the status of install.

i think this file needs to be reorged a little bit to have a step only continue if the previous step succeeded.

because if you didnt upload, you cant unpax. if you didnt unpax, you cant install. if you didnt install, you cant init.

src/actions/InstallationHandler.ts
```diff
@@ -62,11 +62,27 @@ class Installation {
...
+        ProgressStore.set('installation.install', true);
+      }
+
+      let initMvs;
+      if(installation.status){
+        console.log("running zwe init mvs...");
+         initMvs = await this.initMVS(connectionArgs, installationArgs.installationDir);
+        ProgressStore.set('installation.initMVS', initMvs.status);
+      } else {
+        initMvs = {status: true}
```
1000TurquoisePogs: if installMvs = false, initMvs = true?
unconditionally?
i can't think of why this wouldnt be false.

1000TurquoisePogs (changes requested):

src/actions/InstallationHandler.ts
```diff
@@ -62,11 +62,27 @@ class Installation {
...
+        ProgressStore.set('installation.initMVS', initMvs.status);
+      } else {
+        initMvs = {status: false}
+        ProgressStore.set('installation.initMVS', false);
+      }
 
-      return {status: download.status && uploadYaml.status && upload.status && unpax.status && install.status, details: ''};
+      return {status: download.status && uploadYaml.status && upload.status && unpax.status && installation.status && initMvs.status, details: ''};
     } catch (error) {
       return {status: false, details: error.message};
```
1000TurquoisePogs, line 113: This code does not properly present errors to users.

We have a try { } catch { } here, but not all failures go to the catch { } block.

If init mvs fails, there is no exception but the return would say `{status: false, details: ''}`.
There would be no way to differentiate that from a `zwe install` or `extract` step failure.

You should set `details` to what actually failed.
In fact, `initMVS()` is returning a details, so it could be formatted.

timgerstel: This return statement existed prior to this PR, I'd be happy to fix it but I think it should be a separate ticket to enhance the error detailing of this stage @1000TurquoisePogs

DivergentEuropeans: @timgerstel @1000TurquoisePogs don't forget, if there are new errors in Zen the user needs to see, use this:
https://github.com/zowe/zen/pull/45/files#diff-6d106a102cd6a61731f4cb784b463952dd7e6c02a2f1ba3df3e4602e9a4e7e45R222

1000TurquoisePogs (approved):
It is an improvement so I will merge it but there still are situations that need to be caught:

image

If you run it twice, because "--allow-overwrite" is not used, technically `init` does nothing the second time.
It prints out warnings and then says success.
Those warnings should be propagated to the UI, which can be done in a follow-up PR.

PR #92, opened 2023-12-20
Skip action

sakshibobade21:
## Proposed changes

This PR addresses Issue: [*Link to Github issue within https://github.com/zowe/zen/issues* if any]
1) Placement of the buttons for viewing yaml, JCL, and output. I updated this from the one in my previous PR. After carefully considering functionality and code usability, I decided to have these buttons within each stage of initialization instead of the stepper.
2) Incorporating the skip button into the stages in the initialization step. Disabling the skip button if the step is initialized.
3) In the same PR, I also added the bugfix for the regression where the updates in the form of the security step did not persist after moving to another step. 
4) Some Code cleanup.

This PR depends upon the following PRs:

## Type of change
Please delete options that are not relevant.
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Change in a documentation
- [ ] Refactor the code 
- [ ] Chore, repository cleanup, updates the dependencies.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## PR Checklist
Please delete options that are not relevant.
- [ ] If the changes in this PR are meant for the next release / mainline, this PR targets a "staging" branch.
- [ ] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zen/blob/v2.x/staging/CONTRIBUTING.md))
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] New and existing unit tests pass locally with my changes
- [ ] video or image is included if visual changes are made

- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works, or describe a test method below

PR #93, opened 2023-12-30
Add small UI enhancements + fix typos & run-time issue

DivergentEuropeans:
Once we merge this PR, @sakshibobade21's [PR ](https://github.com/zowe/zen/pull/92) should be merged into v2.x/staging

The fixes are too many to count, because there's plenty of small ones. But here's the summary:

- Update Sakshi's 'skip-action' branch with 'v2.x/staging' (Tim's PR). I tried my best to handle the merge conflicts, as some things weren't trivial to discern due to code changes in the same place. But I manually combined both in some places. 

This is the file that needs manual review of those code conflict merges:
zen\src\renderer\components\configuration-wizard\Wizard.tsx

- Added depth to Tim's Stepper UI. Somewhat experimental, intentionally it's the only thing that has depth in Zen. The rest is flat and this makes the whole app look closer to a "filing cabinet" which is a cool look imo for an installation wizard

image

- Added a little life to the buttons
a) some main UI buttons that were off center are center
b) some menu buttons have new icons/new colors
c) when the "Continue" button is available, the "Skip" button becomes hidden (the screenshot shows all buttons for convenience)

image

- There was a small run time issue https://github.com/zowe/zen/pull/92#discussion_r1438153329 , this PR patches it

- Across the board spelling/grammar fixes, or making wording more natural
most changes in:
src/renderer/components/stages/connection/Connection.tsx

menu button changes in:
stepper.tsx


src/renderer/components/stages/Planning.tsx
```diff
@@ -275,17 +294,18 @@ const Planning = () => {
...
-        details.spaceAvailableMb = dfOut.match(/\d+\/\d+/g)[0].split('/')[0];
-        // FIXME: Space requirement is made up, Zowe 2.9.0 convenience build is 515Mb and growing per version. Make it double for extracted files.
-        if (parseInt(details.spaceAvailableMb, 10) < requiredSpace) { 
-          details.error = details.error + `Not enough space, you need at least ${requiredSpace}MB; `;
-        }
-      } catch (error) {
-        details.error = details.error + `Can't check space available; `;
-        console.warn(res[2].details);
-      }
+      //Do not check space because space on ZFS is dynamic. you can have more space than USS thinks.
```
Martin-Zeithaml, line 297: This is true for recommended allocation, which has secondary quantity specified and the zFS is mounted as `AGGRGROW`.  But is is possible to have zFS with "fixed size", however I don't think customers will be using this option. In the future, if we would like to be perfect, we can alter this to detect "fixed size".

src/renderer/components/stages/connection/Connection.tsx
```diff
@@ -163,7 +163,7 @@ const FTPConnectionForm = () => {
           variant="standard"
           helperText={<span style={{display: 'flex', margin: 0}}>
               <img style={{width: '12px', height: '16px', paddingRight: '8px'}} src={secureIcon} alt="secure"/> 
-              <span>We keep your password only for the current session</span>
+              <span>Your password is securely stored for only the current session.</span>
```
Martin-Zeithaml, line 166: I am not English native speaker, but _for only_ does not sound right to me...

DivergentEuropeans: @Martin-Zeithaml Both are grammatically correct - I'm nearly certain. I probably changed that little segment without meaning to, I really wanted to add "securely stored" because I wasn't able to discern/copy/paste out the password credentials

DivergentEuropeans: @Martin-Zeithaml I also just noticed I missed a whole bunch of other typos, most of them by ourselves, so thanks for making me look here again 😃

src/renderer/components/stages/Planning.tsx
```diff
@@ -352,12 +373,12 @@ Please customize job statement below to match your system requirements.
                 id="installation-input"
                 required
                 style={{marginLeft: 0}}
-                label="Installation location (Runtime Directory)"
+                label="Run-time Directory (or installation location)"
```
Martin-Zeithaml, line 376: Docs.zowe.org is referring this directory as runtime (without dash). I think, we should be consistent.

src/renderer/components/stages/connection/Connection.tsx
DivergentEuropeans: Live, long, and prosper is definitely staying

PR #94, opened 2024-01-03
SubStepper UI proposal

skurnevich:
Proposal to have a simpler substepper UI with a circular progress bar instead of separate panel 

Screenshot 2024-01-03 at 15 50 24

### DIFF
--- a/src/renderer/components/common/Stepper.tsx
+++ b/src/renderer/components/common/Stepper.tsx
@@ -14,15 +14,63 @@ import Stepper from '@mui/material/Stepper';
 import Step from '@mui/material/Step';
 import StepLabel from '@mui/material/StepLabel';
 import Button from '@mui/material/Button';
+import Check from '@mui/icons-material/Check';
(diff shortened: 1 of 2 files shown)


timgerstel:
I actually like this a lot, I will run this by the team @skurnevich

skurnevich:
Hi @timgerstel , I actually did that before I noticed that your and Lenny's [PR](https://github.com/zowe/zen/pull/93) also has UI fixes for the stepper. 
I suppose your PR should be merged first, I'm going to review and approve it today. But then there will be a merge conflict, anyway, we can discuss it on the webui squad call tomorrow.

DivergentEuropeans:
@skurnevich Interesting design for sure! 👍  It's minimalist & looks good. I'm a bit biased towards my design though because mine shows more info: you can see secondary steps ahead of you. In this design, you'll need to look elsewhere on the page. So we sacrifice a little vertical real-estate to show users the fuller picture at a glance vs not doing that, which I think is a good sacrifice

PR #113, opened 2024-01-30
[WIP] Remove ConfigService.ts

timgerstel:
## Proposed changes
Previously, there was two services performing the identical task of maintaining the Zowe yaml config. The original ConfigurationStore.ts performed all the operations necessary for Zen, and then ConfigService.ts was added after.

The code base then became scattered with references to two different yaml objects from the two separate services, both of which were being updated independently each time. In short, it was a mess.

This PR removes ConfigService.ts.

## Type of change
- [X] Refactor the code 

## PR Checklist


src/renderer/components/common/EditorDialog.tsx
```diff
@@ -34,24 +33,25 @@ const EditorDialog = ({contentType, isEditorVisible, toggleEditorVisibility, onC
...
   const [editorContent, setEditorContent] = useState(content ? content : '');
   const [isSchemaValid, setIsSchemaValid] = useState(true);
   const [schemaError, setSchemaError] = useState('');
   const fileInputRef = useRef(null);
   let initZoweConfig: any;
 
-  if(contentType == 'yaml') {
-    initZoweConfig = getZoweConfig();
-  }
+
```
DivergentEuropeans, line 44: Can you add a comment why you did this/left it commented

src/renderer/components/common/EditorDialog.tsx
```diff
@@ -74,15 +74,15 @@ const EditorDialog = ({contentType, isEditorVisible, toggleEditorVisibility, onC
     }
 
     if(newCode && (newCode == "\n" || newCode == "")) {
-      setZoweConfig("");
+      // setZoweConfig("");
```
DivergentEuropeans, line 77: Can you add a comment why you did this/left it commented

DivergentEuropeans (approved):

src/renderer/components/common/EditorDialog.tsx
timgerstel: initZoweConfig was never called anywhere, its just dead code from staging that should be 100% removed

src/renderer/components/common/EditorDialog.tsx
timgerstel: Dead logic from staging that should be 100% removed

PR #79, opened 2023-10-28
Add verifyCertificates option and 'Run zwe init certificates' button to Certificate Stage

timgerstel:
## Proposed changes
This pull request implements the following changes:

- Enhance Certificate stage of Zen
- Add UI elements to set zowe.verifyCertificates
- Add UI button to run 'zwe init certificates'

image

image

This PR depends upon the following PRs:

## Type of change
Please delete options that are not relevant.
- [X] New feature (non-breaking change which adds functionality)

## PR Checklist
Please delete options that are not relevant.
- [ ] If the changes in this PR are meant for the next release / mainline, this PR targets a "staging" branch.
- [ ] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zen/blob/v2.x/staging/CONTRIBUTING.md))
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] New and existing unit tests pass locally with my changes
- [ ] video or image is included if visual changes are made

- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works, or describe a test method below

PR #114, opened 2024-01-30
Remove ConfigService.ts, Enahance Certifcate Stage

timgerstel:
## Proposed changes
This PR combines these two PRs:
- https://github.com/zowe/zen/pull/113
- https://github.com/zowe/zen/pull/79

## Type of change
- [ ] Refactor the code 

## PR Checklist


src/renderer/preload.ts
```diff
@@ -64,8 +64,11 @@ contextBridge.exposeInMainWorld('electron', {
     checkDirOrCreate(connectionArgs: IIpcConnectionArgs, location: string) {
       return ipcRenderer.invoke("check-dir-or-create", connectionArgs, location);
     },
-    installButtonOnClick(connectionArgs: IIpcConnectionArgs, installationArgs: {installDir: string, javaHome: string, nodeHome: string, installationType: string, userUploadedPaxPath: string, smpeDir: string}, version: string, zoweConfig: any) {
-      return ipcRenderer.invoke("install-mvs", connectionArgs, installationArgs, version, zoweConfig);
+    installButtonOnClick(connectionArgs: IIpcConnectionArgs, installationArgs: {installDir: string, javaHome: string, nodeHome: string, installationType: string, userUploadedPaxPath: string, smpeDir: string}, version: string, zoweConfig: any, skipDownload: boolean) {
```
DivergentEuropeans, line 67: what is this skipDownload stuff for? If it's for smpe, why don't we just check if smpeDir exists? Wasn't that the reason why we made that variable in the first place instead of setting installationDir = smpeDir and deleting smpeDir entirely

Also, it looks like it's never actually set any where, except 1 default
`window.electron.ipcRenderer.installButtonOnClick(connectionArgs, installationArgs, version, yaml, skipDownload ?? false)`

timgerstel: It was for QA testing, not a prod solution. It was so @struga0258 can skip the pax download process.

timgerstel: And its set here: https://github.com/zowe/zen/pull/114/files#diff-876487e16434db86af62a71372589b9b6ee12ec10cc731bc989c96eb78d3b17fR196

DivergentEuropeans: Oh I see. Yes, we need a better solution for the smpe step

DivergentEuropeans (changes requested):
we need a better solution for the smpe step/QA testing hack

src/actions/InstallActions.ts
```diff
@@ -35,11 +35,16 @@ export class InstallActions {
     }
   }
 
+  runZweInitCertificates(connectionArgs: IIpcConnectionArgs, 
+    installationArgs: {installationDir: string}, zoweConfig: any){
+      return new FTPInstallation().initCertificates(connectionArgs, installationArgs.installationDir, zoweConfig)
```
DivergentEuropeans: This doesn't seem right. It seems like 
1. initCertificates( ) should should belong in `Installation `class, with initSecurity & the others +
initCertificates should be made public
2. change this to
`return this.strategy.initCertificates(connectionArgs, installationArgs.installationDir, zoweConfig);`

src/actions/InstallationHandler.ts
```diff
@@ -39,43 +40,46 @@ class Installation {
...
+        unpax = await this.unpax(connectionArgs, installationArgs.installationDir); 
+        ProgressStore.set('installation.unpax', unpax.status);
 
-      if(!unpax.status){
-        return {status: false, details: `Error unpaxing Zowe archive: ${unpax.details}`};
+        if(!unpax.status){
+          return {status: false, details: `Error unpaxing Zowe archive: ${unpax.details}`};
+        }
+      } else {
+        download = upload = unpax = {status: true, details : ''}
```
DivergentEuropeans, line 82: Never seen that b4 for some reason, cool 👍

src/actions/InstallationHandler.ts
```diff
@@ -251,6 +255,28 @@ export class FTPInstallation extends Installation {
...
+      } 
+    });
+    ProgressStore.set('certificate.writeYaml', true);
+    console.log("uploading yaml...");
+    const uploadYaml = await this.uploadYaml(connectionArgs, installDir);
+    if(!uploadYaml.status){
+      return ProgressStore.set('certificate.uploadYaml', false);;
+    }
+    ProgressStore.set('certificate.uploadYaml', uploadYaml.status);
+    const script = `cd ${installDir}/runtime/bin;\n./zwe init certificate --update-config -c ${installDir}/zowe.yaml`;
```
DivergentEuropeans: you no longer need "\n" as it's handled automatically however we do want to make sure we're using "&&" instead of ";"

src/main/index.ts
```diff
@@ -136,8 +136,13 @@ const createWindow = (): void => {
...
   })
 
-  ipcMain.handle('install-mvs', async (_event, connectionArgs, installationArgs, version, zoweConfig) => {
-    const res = await installActions.runInstallation(connectionArgs, installationArgs, version, zoweConfig);
+  ipcMain.handle('install-mvs', async (event, connectionArgs, installationArgs, version, zoweConfig, skipDownload) => {
+    const res = await installActions.runInstallation(connectionArgs, installationArgs, version, zoweConfig, skipDownload);
+    return res;
+  });
+
+  ipcMain.handle('zwe-init-certificates', async (event, connectionArgs, installationArgs, zoweConfig) => {
```
DivergentEuropeans: why does this have `zwe-init-certificates` but below is `init-apf `?

src/actions/InstallActions.ts
DivergentEuropeans: Also, the commands below don't include the `Zwe `part from method name so I guess we could remove that for organization's sake

DivergentEuropeans (changes requested):

DivergentEuropeans (approved):
Changes addressed

PR #125, opened 2024-03-01
Editor store + make Zen smarter about errors (stricter error checking, show more errors, human readable)

DivergentEuropeans:
- prints more errors (zwe init + zwe install + jcl job submission)

TODO: JCL job submission error may be long. This error should go into an Editor too. We just don't show the Editor button in the JCL submission step, so I'll think about it more 

- change Zen to use `&&` instead of `;` so multiple commands don't fail at once, causing potential harm
- makes it so Zen doesn't let you proceed normally if zwe install, zwe init apfauth, or zwe init security fails

we more correctly check for if an error occurred 
(via text search-zwe z/OS Unix failure trigger instead of undependable status/rc that is sometimes `0` when it shouldn't be `0`)

- we more correctly show error to the user, via opening the filtered, relevant portion, via the Editor
- Because we use the Editor, to store outputs sent to/received from the Editor and to not be lost after session refresh/close, we made an Editor store
- ^the above also caused me to do some refactoring of the Stores (aka remove code duplication, put common code in a `DefaultStore`, all `Stores `build from the `DefaultStore`)
- minor typo/method renames, for example: 
`ConnectionStore.set` --> `ConnectionStore.setAndValidate` because it goes through schema validation (unchanged behaviour by me, just making things make more fucking sense)

Here's an example of these new smarts together, via a failed Apfauth step:

image

Failed job submission example:

image

**Note: the Editor stuff is not ALL hooked up to the Editor Store. This requires more code changes & can be in a future PR. The only actions right now hooked up to the store are the zwe command/job output.**


src/renderer/components/stages/InitApfAuth.tsx
```diff
@@ -107,22 +112,39 @@ const InitApfAuth = () => {
...
-        dispatch(setInitializationStatus(false));
-        dispatch(setApfAuthStatus(false));
+        apfAuthProceedActions(false);
+        toggleProgress(false);
+        // TODO: Test this
+        //alertEmitter.emit('showAlert', err.toString(), 'error');
         stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = true;
         stages[STAGE_ID].isSkipped = true;
-        console.warn('zwe init apfauth failed');
+        console.warn('zwe init apfauth failed', err);
```
DivergentEuropeans: Add a stringify here if err isn't a string

src/renderer/components/stages/Planning.tsx
```diff
@@ -748,7 +750,7 @@ Please customize the job statement below to match your system requirements.
                       id="zosmf-host"
                       required
                       style={{marginLeft: 0}}
-                      label="z/OSMF Host"
+                      label="z/OSMF host"
```
DivergentEuropeans: change this back to capital case

src/renderer/components/stages/Security.tsx
```diff
@@ -109,22 +108,39 @@ const Security = () => {
...
+        toggleProgress(false);
         clearInterval(timer);
-        dispatch(setNextStepEnabled(false));
-        dispatch(setSecurityStatus(false));
-        dispatch(setInitializationStatus(false));
+        securityProceedActions(false);
         stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = true;
         stages[STAGE_ID].isSkipped = true;
-        console.warn('zwe init security failed');
+        console.warn('zwe init security failed', err);
```
DivergentEuropeans: stringify point here

src/renderer/components/stages/installation/Installation.tsx
```diff
@@ -122,44 +118,53 @@ const Installation = () => {
...
-        }).catch(() => {
-          clearInterval(timer);
-          dispatch(setNextStepEnabled(false));
-          dispatch(setInitializationStatus(false));
-          dispatch(setDatasetInstallationStatus(false));
+          installProceedActions(false);
           stages[stageId].subStages[subStageId].isSkipped = true;
           stages[stageId].isSkipped = true;
-          console.warn('Installation failed');
+          console.warn('Installation failed', err);
```
DivergentEuropeans: stringify?

timgerstel (approved):

DivergentEuropeans (commented):
From my code review w/ @timgerstel

PR #128, opened 2024-03-03
Remove network calls for yaml and schema

timgerstel:
## Proposed changes

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/src/renderer/components/stages/Planning.tsx
+++ b/src/renderer/components/stages/Planning.tsx
@@ -239,6 +239,103 @@ const EXAMPLE_YAML = {
   }
 }
 
+const serverSchema = {
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "$id": "https://zowe.org/schemas/v2/server-common",
+  "title": "Common types",
+  "description": "Configuration types that are common in Zowe and may be referenced by multiple components",
+  "$defs": {
+    "semverVersion": {
+      "$anchor": "zoweSemverVersion",
+      "type": "string",
+      "description": "A semantic version, see https://semver.org/",
+      "pattern": "^[0-9]*\\.[0-9]*\\.[0-9]*(-*[a-zA-Z][0-9a-zA-Z\\-\\.]*)?(\\+[0-9a-zA-Z\\-\\.]*)?$"
+    },
+    "semverRange": {
+      "$anchor": "zoweSemverRange",
+      "type": "string",
+      "description": "A semantic version, see https://semver.org/",
+      "pattern": "^(([\\^\\~\\>\\<]?)|(>=?)|(<=?))[0-9]*\\.[0-9]*\\.[0-9]*(-*[a-zA-Z][0-9a-zA-Z\\-\\.]*)?(\\+[0-9a-zA-Z\\-\\.]*)?$"
+    },
+    "dataset": {
+      "$anchor": "zoweDataset",
+      "type": "string",
+      "description": "A 44-char all caps dotted ZOS name",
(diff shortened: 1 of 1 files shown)


src/renderer/components/stages/Planning.tsx
```diff
@@ -220,24 +1400,16 @@ const Planning = () => {
...
-              delete schema.$defs?.networkSettings?.properties?.server?.properties?.listenAddresses?.items?.ref;
-              schema.$defs.networkSettings.properties.server.properties.listenAddresses.items = serverSchema.$defs.ipv4
-            }
-            dispatch(setSchema(schema));
-          }); 
-        }); 
+        dispatch(setYaml(EXAMPLE_YAML));
+        setLocalYaml((EXAMPLE_YAML));
+        YAML_SCHEMA.properties.zowe.properties.setup.properties.dataset.properties.parmlibMembers.properties.zis = serverSchema.$defs.datasetMember;
+        YAML_SCHEMA.properties.zowe.properties.setup.properties.certificate.properties.pkcs12.properties.directory = serverSchema.$defs.path;
```
DivergentEuropeans: If `YAML_SCHEMA` is a var, that's first set & then redefined here,, and `serverSchema `is a const, can't you just define `YAML_SCHEMA` to be the correct value in the first place? Why set it once, and then set two of its properties?

src/renderer/components/stages/Planning.tsx
```diff
@@ -220,24 +1400,16 @@ const Planning = () => {
...
+        YAML_SCHEMA.$id = serverSchema.$id;
+        if(YAML_SCHEMA.$defs?.networkSettings?.properties?.server?.properties?.listenAddresses?.items){
+          delete YAML_SCHEMA.$defs?.networkSettings?.properties?.server?.properties?.listenAddresses?.items?.ref;
+          YAML_SCHEMA.$defs.networkSettings.properties.server.properties.listenAddresses.items = serverSchema.$defs.ipv4
+        }
```
DivergentEuropeans: If we have predefined `YAML_SCHEMA` , and then set two of its properties, and now we're here, wouldn't we know what that object is set to? So why are we using this conditional when it wouldn't change? If this is a thing where, the schema isn't correct and like in similar code above, you need to prune/trim the schema so the parser doesn't spit out an error - then that's understandable, but at least put into a utility function. Lmk if there's something I'm not understanding here

src/renderer/components/stages/Planning.tsx
DivergentEuropeans: I see it was also done here: https://github.com/zowe/zen/pull/128/files#diff-6d106a102cd6a61731f4cb784b463952dd7e6c02a2f1ba3df3e4602e9a4e7e45R1386

if it's like a "change my schema object to not break schema parser", then you could put it into a function. Same as the thing below

DivergentEuropeans (approved):
This otherwise looks good

src/renderer/components/stages/Planning.tsx
timgerstel: I mean yeah that logic was from the original Zen code, it could be refactored but i left it for some reason

src/renderer/components/stages/Planning.tsx
timgerstel: Again from the original Zen code

timgerstel:
Merged as part of https://github.com/zowe/zen/pull/114

DivergentEuropeans (approved):
This PR removes some magic, which is good

PR #146, opened 2024-03-20
Retrieve example-zowe.yaml, zowe-yaml-schema.json and server-common.json Schemas from Zowe Runtime Dir (for upload/download install types)

timgerstel:
## Proposed changes

## Type of change
- [ ] Refactor the code 

## PR Checklist


src/actions/InstallationHandler.ts
```diff
@@ -82,6 +84,100 @@ class Installation {
         download = upload = unpax = {status: true, details : ''}
       }
 
+      let zoweRuntimePath = installationArgs.installationType === "smpe" ? installationArgs.installationDir : installationArgs.installationDir + "/runtime";
+      let readPaxYamlAndSchema = await this.readExYamlAndSchema(connectionArgs, zoweRuntimePath);
```
timgerstel: readExYaml.... -> readExampleYaml...()

src/actions/InstallationHandler.ts
```diff
@@ -81,29 +85,14 @@ class Installation {
...
-      }
-      ProgressStore.set('installation.upload', upload.status);
-
-      if(!upload.status){
-        return {status: false, details: `Error uploading pax: ${upload.details}`};
-      }
-
       unpax = {status: false, details: ""};
       if (!SMPE_INSTALL) {
         console.log("unpaxing...");
```
DivergentEuropeans, line 97: Doesn't this duplicate the code in line 79? Are `!SMPE_INSTALL` and `!skipDownload` mutually exclusive? They both do:
```
console.log("unpaxing...");
        unpax = await this.unpax(connectionArgs, installationArgs.installationDir); 
        ProgressStore.set('installation.unpax', unpax.status);

        if (!unpax.status) {
          return {status: false, details: `Error unpaxing Zowe archive: ${unpax.details}`};
        }
  ```
  Is this intentional? Basically it seems like you correctly removed code, but you didn't remove all the duplication. We wouldn't want to unpax twice if we don't need to. It might be helpful to reorganize everything/add comments in the
  ``` 
  try {
  ....
  } catch (error) {
      return {status: false, details: error.message};
    }
```   
stuff to make it easier to follow

src/actions/InstallationHandler.ts
```diff
@@ -114,6 +103,114 @@ class Installation {
...
+        if(yamlFromPax){
+          try {
+            yamlObj = parse(yamlFromPax);
+            if (currentConfig) {
+              console.log("current config:", JSON.stringify(currentConfig));
+              // console.log("yamlObj: ", JSON.stringify(yamlObj));
+              yamlObj = {...currentConfig, ...yamlObj}
+              console.log("merged yamlObj: ", JSON.stringify(yamlObj));
+            }
+            if (yamlObj.zowe.runtimeDirectory === undefined && installationArgs.installationDir) {
```
DivergentEuropeans, line 128: Why are we checking if the yaml object's properties are undefined first. Wouldn't we want to `yamlObj` properties always 100% set to `installationArgs` stuff (assuming `installationArgs` stuff is user input stuff? or is it sometimes default Zen stuff excluding user input?)

src/actions/InstallationHandler.ts
```diff
@@ -317,6 +414,16 @@ export class FTPInstallation extends Installation {
     return {status: result.rc === 0, details: result.jobOutput}
   }
 
+  async readExYamlAndSchema(connectionArgs: IIpcConnectionArgs, installDir: string){
```
DivergentEuropeans: here too

src/renderer/components/stages/Planning.tsx
```diff
@@ -1448,24 +255,23 @@ const Planning = () => {
...
     e.preventDefault();
     setValidationDetails({...validationDetails, error: ''});
-    setPlanningValidationDetailsState({...validationDetails, error: ''});
-    if (!localYaml?.java?.home || !localYaml?.node?.home || !localYaml?.zowe?.runtimeDirectory) {
-      console.warn('Please fill in all values');
-      alertEmitter.emit('showAlert', 'Please fill in all values', 'error');
-      //showAlert('Please fill in all values', 'success', 5000);
-      return;
-    }
+    // if (!localYaml?.java?.home || !localYaml?.node?.home || !localYaml?.zowe?.runtimeDirectory) {
```
DivergentEuropeans: Do we not want to check if both Java and Node home are set? What happens if user has them empty

src/renderer/components/stages/Planning.tsx
```diff
@@ -1855,12 +682,14 @@ Please customize the job statement below to match your system requirements.
                       style={{marginLeft: 0}}
                       label="z/OSMF Host"
                       variant="standard"
-                      value={localYaml?.zOSMF?.host || installationArgs.zosmfHost}
+                      value={localYaml?.zOSMF?.host || installationArgs.zosmfHost || "dvipa.my-company.com"}
```
DivergentEuropeans: Shouldn't this come from fallback YAML instead of hardcode?

src/renderer/components/stages/Planning.tsx
```diff
@@ -1701,13 +514,15 @@ Please customize the job statement below to match your system requirements.
                 style={{marginLeft: 0}}
                 label="Extensions Directory"
                 variant="standard"
-                value={localYaml?.zowe?.extensionDirectory || installationArgs.extensionDir}
+                value={localYaml?.zowe?.extensionDirectory || installationArgs.extensionDir || "/global/zowe/extensions"}
```
DivergentEuropeans: Shouldn't this come from fallback YAML instead of hardcode?

src/renderer/components/stages/Planning.tsx
```diff
@@ -1681,13 +492,15 @@ Please customize the job statement below to match your system requirements.
                 style={{marginLeft: 0}}
                 label="Log Directory"
                 variant="standard"
-                value={localYaml?.zowe?.logDirectory || installationArgs.logDir}
+                value={localYaml?.zowe?.logDirectory || installationArgs.logDir || "/global/zowe/logs"}
```
DivergentEuropeans: Shouldn't this come from fallback YAML instead of hardcode?

src/renderer/components/stages/Planning.tsx
```diff
@@ -1661,13 +470,15 @@ Please customize the job statement below to match your system requirements.
                 style={{marginLeft: 0}}
                 label="Workspace Directory"
                 variant="standard"
-                value={localYaml?.zowe?.workspaceDirectory || installationArgs.workspaceDir}
+                value={localYaml?.zowe?.workspaceDirectory || installationArgs.workspaceDir || "/global/zowe/workspace"}
```
DivergentEuropeans: Shouldn't this come from fallback YAML instead of hardcode?

DivergentEuropeans (commented):
I'm sure this works and the basic logic looks good, in how objects get passed around, but due to the flow of data priority of user input (installationArgs) vs. schema'd user input (localYaml) vs. hardcoded defaults being set in multiple steps across multiple places, it could be more clear. I'm not exactly how to help with this atm but some good comments would be nice. At the least like, why are we setting hardcoded defaults as a "just in case" fallback in the UI code, 
```
value={localYaml?.zowe?.workspaceDirectory || installationArgs.workspaceDir || "/global/zowe/workspace"}
```
when we could be doing it further up? Unless there is some edgecase where both localYaml and installationArgs are not set? Then 
a) why does this edgecase happen and 
b) we need a comment why we need the hardcode

DivergentEuropeans:
Also if reorging the flow of data is too out of scope for this PR, then understand, but we still need better TODOs

src/actions/InstallationHandler.ts
timgerstel: Skip download isn't even part of the UI right now and I dont think we have plans for it for the first release. I only added skipDownload for QA

src/actions/InstallationHandler.ts
timgerstel: You are correct for this one, good catch! I think i made some late night changes and did this by mistake. Will fix.

src/actions/InstallationHandler.ts
timgerstel: Fixed

src/renderer/components/stages/Planning.tsx
timgerstel: I had commented this out because it was always showing the error, even when java.home, node.home and zowe.runtimeDirectory were filled in. I will uncomment it and test if this still happens because this check should occur

timgerstel:
> I'm sure this works and the basic logic looks good, in how objects get passed around, but due to the flow of data priority of user input (installationArgs) vs. schema'd user input (localYaml) vs. hardcoded defaults being set in multiple steps across multiple places, it could be more clear. I'm not exactly how to help with this atm but some good comments would be nice. At the least like, why are we setting hardcoded defaults as a "just in case" fallback in the UI code,
> 
> ```
> value={localYaml?.zowe?.workspaceDirectory || installationArgs.workspaceDir || "/global/zowe/workspace"}
> ```
> 
> when we could be doing it further up? Unless there is some edgecase where both localYaml and installationArgs are not set? Then a) why does this edgecase happen and b) we need a comment why we need the hardcode

The reason this edgecase happens:

On the very first run of Zen, there will be no YAML or installationArgs. It's that simple.

The planning stage also doesn't use json-forms, so defaults from the schema arent fetched.

I did change it to use the fallback yaml though.

src/actions/InstallationHandler.ts
timgerstel: I actually added these lines back because line 126, you can see the currentConfig is the first thing that is merged into the yaml. Therefore, if the user has already entered zowe.runtimeDirectory in their current config we do not want to overwrite it with intsallationArgs (which may be null after an app restart anyway).

The undefined checks are necessary.

DivergentEuropeans (approved):

DivergentEuropeans:
Changes addressed

PR #148, opened 2024-04-03
(WIP) Editor store + make Zen smarter about errors (stricter error checking, show more errors, human readable)

DivergentEuropeans:
bugs to review possibly in this PR:

1.When we intilaize MVS datasets and click on view&Submit job button then its breaks the zen, open the blank screen - Recent Merge
2.Updating values in the planning stage does not update the values in zowe.yaml, and returning to the planning stage shows default values. It does not save the values even if we close Zen and reopen it. - Recent Merge
3. Additionally, updating values in the YAML file does not update the values in the background within the fields - Recent Merge
4.Also even after filling all values at planning stage still gives error please fill all values- Recent Merge

features of PR:

- prints more errors (zwe init + zwe install + jcl job submission)

TODO: JCL job submission error may be long. This error should go into an Editor too. We just don't show the Editor button in the JCL submission step, so I'll think about it more 

- change Zen to use `&&` instead of `;` so multiple commands don't fail at once, causing potential harm
- makes it so Zen doesn't let you proceed normally if zwe install, zwe init apfauth, or zwe init security fails

we more correctly check for if an error occurred 
(via text search-zwe z/OS Unix failure trigger instead of undependable status/rc that is sometimes `0` when it shouldn't be `0`)

- we more correctly show error to the user, via opening the filtered, relevant portion, via the Editor
- Because we use the Editor, to store outputs sent to/received from the Editor and to not be lost after session refresh/close, we made an Editor store
- ^the above also caused me to do some refactoring of the Stores (aka remove code duplication, put common code in a `DefaultStore`, all `Stores `build from the `DefaultStore`)
- minor typo/method renames, for example: 
`ConnectionStore.set` --> `ConnectionStore.setAndValidate` because it goes through schema validation (unchanged behaviour by me, just making things make more fucking sense)

Here's an example of these new smarts together, via a failed Apfauth step:

image

Failed job submission example:

image

**Note: the Editor stuff is not ALL hooked up to the Editor Store. This requires more code changes & can be in a future PR. The only actions right now hooked up to the store are the zwe command/job output.**


DivergentEuropeans:
Duplicate

PR #155, opened 2024-05-08
State Management for the Active Stage

sakshibobade21:
## Proposed changes

This PR manages the state for the active stage in multiple sessions

## Type of change
- [ ] Refactor the code 

## PR Checklist


timgerstel (changes requested):
image
Original resume button should be used instead, and should persist between app restarts

src/renderer/components/configuration-wizard/Wizard.tsx
```diff
@@ -26,19 +26,19 @@ import Networking from '../stages/Networking';
...
-  {id: 0, label: 'Connection', component: <Connection/>, hasJCL: false, isSkippable: false, isSkipped: false, hasOutput: false, steps: 1, nextButton: 'Continue'},
-  {id: 1, label: 'Planning', component: <Planning/>, hasJCL: false, isSkippable: false, isSkipped: false, hasOutput: true, steps: 3, nextButton: 'Continue to Installation Options'},
-  {id: 2, label: 'Installation Type', component: <InstallationType/>, hasJCL: false, isSkippable: false, isSkipped: false, hasOutput: false, steps: 1, nextButton: 'Continue to Components Installation'},
+  {id: 0, label: 'Connection', component: <Connection/>, hasJCL: false, isSkippable: false, isSkipped: false, hasOutput: false, steps: 1, nextButton: 'Continue', statusKey: 'connectionStatus'},
+  {id: 1, label: 'Planning', component: <Planning/>, hasJCL: false, isSkippable: false, isSkipped: false, hasOutput: true, steps: 3, nextButton: 'Continue to Installation Options', statusKey: 'planningStatus'},
+  {id: 2, label: 'Installation Type', component: <InstallationType/>, hasJCL: false, isSkippable: false, isSkipped: false, hasOutput: false, steps: 1, nextButton: 'Continue to Components Installation', statusKey: 'installationTypeStatus'},
```
timgerstel, line 31: Remove statusKey and use the label instead for the getProgress() arg, the objects in this array are already quite messy

src/renderer/components/stages/connection/Connection.tsx
```diff
@@ -162,6 +153,26 @@ const FTPConnectionForm = () => {
...
+  const resumeProgress = () => {
+    window.electron.ipcRenderer.getConfig().then((res: IResponse) => {
+      if (res.status) {
+        dispatch(setYaml(res.details.config));
+        const schema = res.details.schema;
+        dispatch(setSchema(schema));
+      } else {
+        dispatch(setYaml(EXAMPLE_YAML));
+        dispatch(setSchema(YAML_SCHEMA));
+        window.electron.ipcRenderer.setConfig(EXAMPLE_YAML).then((res: IResponse) => {
```
timgerstel, line 166: Settings config and schema here will conflict with this PR:

https://github.com/zowe/zen/pull/146

These ipcRenderer calls as well as constants.ts should be removed.

Was there a specific reason you needed to set these here?

src/renderer/components/common/Stepper.tsx
```diff
@@ -165,17 +167,17 @@ export default function HorizontalLinearStepper({stages, initialization}:{stages
     setActiveStep(newActiveStep);
     const newSubStep = isSubStep ? subStepIndex : 0;
     setActiveSubStep(newSubStep);
-  };
+  }
 
   const getStepIcon = (error: any, stageId: number, isSubStep?: boolean, subStepId?: number) => {
     
+    if (!error || (isSubStep && getProgress(stages[stageId].subStages[subStepId].statusKey)) || (!isSubStep && getProgress(stages[stageId].statusKey))) {
```
timgerstel, line 177: Remove statusKey, use label

sakshibobade21:
The functionality of the resume button operates as intended under two different scenarios:
When the user seeks to resume progress within the same session.
When the user intends to resume progress in a subsequent session, such as after restarting the application.
In the second scenario, adding the resume button on the home screen would be impractical, as resuming progress should only be permissible after establishing a successful connection.
Therefore, in this case, the resume button is appropriately placed within the connection stage and becomes visible only upon successful connection.
Conversely, in the first scenario where a successful connection is already established, the resume button is on the home screen, because on clicking "Save and close" you are redirected to the home screen.
Lmk if it looks good to you all. I am open for suggestions.

src/renderer/components/stages/connection/Connection.tsx
sakshibobade21: This code is used to set the schema and the yaml. We need this because, when we resume to any stage say "Certificate" directly after connection we need to have the schema.

src/renderer/components/configuration-wizard/Wizard.tsx
sakshibobade21: The function getProgress() retrieves an object from local storage. Keys in the local storage are stored using camel case conventions. The labels are displayed in the UI therefore they cannot be in camel case.
I prefer to maintain camel case conventions for the keys.

timgerstel:
Two very very small things:

1) On the very first run of Zen, the "Saved Installation" tile should probably be hidden since there can't be a previous installation: image.

2) The wording of "Validate Credentials & Continue to the Last Active Stage." is a little awkward, perhaps we should reuse "Validate Credentials & Resume" as the tooltip for the home stage resume button.

Other than that I tested this PR and it works as expected. If QA has also tested this feature, then this PR looks good to me!

timgerstel (approved):

PR #166, opened 2024-06-05
Fix macos build

1000TurquoisePogs:
Macos isnt building because of 
1) appdmg failing to install which is due to python.
2) the github builder now builds macos in arm64 edition instead of x64. that's fine, but it needs a workflow update.


sakshibobade21 (approved):
Look good to me. Thanks Sean.

1000TurquoisePogs:
DCO bot is hung and preventing merge. I am forcing this through after confirming all DCO is correct.

PR #169, opened 2024-06-05
Set of hotfixes

skurnevich:
- Corrected missing method setConfigByKey
- Fixed Networking panel losing focus on components port input
- Added missing keys for generated children
- Overlay z-index
- Removed CLI check and summary on Planning as not implemented


timgerstel (approved):

DivergentEuropeans:
@skurnevich 

> Removed CLI check and summary on Planning as not implemented

I still get [11152:0606/195526.590:INFO:CONSOLE(96)] "No Zowe CLI found on local machine", source: webpack-internal:///./src/renderer/components/Home.tsx (96) 

> Fixed Networking tab losing focus on components port input

I trust you but can't replicate. You saying tabbing doesn't work over ports? I can tab there fine

skurnevich:
@DivergentEuropeans 
I didn't remove all CLI code, just removed check for it so it always acts like there is no CLI installed, so you see this message.
As for Networking, tabulation was working fine, but if you try to type in the port field it loses focus after each digit and you need to click it back (or use tab). 
Basically the fix is hard to spot in the PR code because it is one line fix, line 80, key for the input textbox has port value in it, so with every new digit react thinks it is a different input element and loses focus.

PR #172, opened 2024-06-07
Stage Persistence for Networking and Launch Config Stage

sakshibobade21:
## Proposed changes

This PR addresses Issue: Persist the state for the Networking and the Launch Config Stage

## Type of change
- [ ] Refactor the code 

## PR Checklist


src/renderer/components/common/MonacoEditor.tsx
```diff
@@ -11,16 +11,22 @@
 import React, { useEffect, useRef, useState } from 'react';
 import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
 import { load } from 'js-yaml';
+import { diff } from 'deep-diff';
```
timgerstel: This library is unmaintained so it should definitely not be added. I think there's probably a better solution to the undefined "externalDomains" issue. For now if you could please remove the fix for externalDomains, I'll work on that and then I think the rest of this PR is ok so I will approve

timgerstel (approved):

src/renderer/components/common/MonacoEditor.tsx
timgerstel: We also didn't need to check the diff of the yaml on any other stages so unless there's an exception on this stage that I missed I don't think we'll need to even do a diff comparison

timgerstel (changes requested):
Meant to request changes, approved by accident

src/renderer/components/common/MonacoEditor.tsx
timgerstel: Fix here: https://github.com/zowe/zen/pull/175

sakshibobade21: I am checking the diff because if in the yaml attributes from the networking stage is updated then we should mark the networking stage as completed as discussed with @DivergentEuropeans

sakshibobade21: If we update anything in the form then the networking stage is mark as completed.
But what if the form is not updated, but instead the networking attributes in the yaml is updated?
In this case we need to check the diff of the yaml.

timgerstel: There are much easier ways to check if the yaml has been changed or updating than by taking the diff. Diffs can be an expensive operation depending on what is being compared. Also as I stated before, the deep-diff library you added hasn't been maintained or updated in 6 years.

In my PR that fixes the networking stage, I set the stage to complete if the users yaml successfully uploads after clicking "save YAML to z/OS button" on the Networking and Launch Config stages: https://github.com/zowe/zen/pull/175/files#diff-9b9ad16497a71852bcfc086ef0d71306ec3d8ad6760ee1a2d731fc60c8ea1ee9R530


I will ask again that you please remove deep-diff and anything related. Thank you.

sakshibobade21: We can definitely check if the yaml has been updated or not easily.
But I specifically wanted to check if the components or the launch config objects has been updated.

sakshibobade21: I can remove the deep-diff and the associated changes but then will we need a separate ticket to change the networking and launch config stage status to completed if the yaml is upadting the config rekated to these srage?

timgerstel: I handled that in my PR if you re-read my last reply

sakshibobade21: I of course read your reply. 

I am saying this because what we are discussing now contradicts what we discussed in the grooming.

You say that you update the stage status to completion when the user's yaml successfully uploads after clicking "save YAML to z/OS button". That makes sense to me.

But what if user is on installation stage and updates the component section in the yaml? Will the networking stage still show the completion mark? According to the discussion and I believe, it should. 

Right @DivergentEuropeans ?

timgerstel: I must have missed that during grooming because I would have disagreed with that.

I do not think changing the YAML should be setting the stage status of ANY stage.

sakshibobade21: I precisely remember that being discussed without anybody disagreeing.

timgerstel: I'm sure you're correct and I probably missed it, that's on me

sakshibobade21: This was discussed for networking and the launch config stage.

src/renderer/components/common/Stepper.tsx
```diff
@@ -216,6 +216,28 @@ export default function HorizontalLinearStepper({stages, initialization}:{stages
 
   const isNextStepEnabled = useAppSelector(selectNextStepEnabled);
 
+  const skipButtonDisabled = (stages: any, activeStep: number, activeSubStep: number) => {
+    if(stages[activeStep]) {
+      if(stages[activeStep].isSkippable && !stages[activeStep].subStages) {
+        if(getProgress(stages[activeStep].statusKey)) {
```
timgerstel, line 222: Looks like this might make the "Unpax" stage unskippable. We want users to be able to skip the download step if they have already performed that process, even if Zen doesn't recognize/remember the status of the Unpax stage

src/renderer/components/common/MonacoEditor.tsx
sakshibobade21: We can definitely update the stage to completion only when the user clicks "SAVE YAML TO ZOS".
I am just saying, I implemented what we discussed in the grooming.

src/renderer/components/common/Stepper.tsx
sakshibobade21: If you test the PR unpax stage can be skipped

sakshibobade21:
Closing this Pull Request. Created a new one.

src/renderer/components/common/MonacoEditor.tsx
DivergentEuropeans: Oops, arriving to this conversation late @sakshibobade21 and @timgerstel - I was on PTO

So, because there was no button in the Networking stage, the only way to see if this stage was edited OR "skipped" is to see if any user input at all has happened. That was the previous story, discussed during a team meet so that's correct. That could be checked via diff but it seems @timgerstel gave some good reasons for not having this library. Also, like @timgerstel said, since we now have a button, let's instead link status to this button instead. Technically you never needed to do a diff in the first place @sakshibobade21 you could have just kept a listener on if any of the UI elements have been changed. Doing diff logic would be hard, because like what if you do bulk copy-pastes and stuff like that, or change Networking section while not being in Networking stage, then you need any kind of diff checking to be global - it would just get heavy quickly

PR #174, opened 2024-06-07
Parsing issues

skurnevich:
That solves the parsing issue on my side, however it need to be tested on different environments, If it works without that fix for someone then it will probably break things for them. 

As a side note, I find this approach of parsing output of `cat` command rather odd, why don’t we use FileTransfer to just download needed files?

### DIFF
--- a/src/actions/InstallationHandler.ts
+++ b/src/actions/InstallationHandler.ts
@@ -141,9 +141,9 @@ class Installation {
       let parsedSchema = false, parsedYaml = false;
       if(readPaxYamlAndSchema.details.yaml){
         const parseExampleYamlFromPax = function(catPath: string){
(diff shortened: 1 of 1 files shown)


DivergentEuropeans:
@skurnevich is this supposed to fix

16:36:23.661 > error setting schema from pax: SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at FTPInstallation.downloadUnpax (C:\Program Files\Git\zen\.webpack\main\index.js:25847:47)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async C:\Program Files\Git\zen\.webpack\main\index.js:26511:21
    at async WebContents.<anonymous> (node:electron/js2c/browser_init:2:89579)

?

DivergentEuropeans:
It doesn't fix the above error 4 me

skurnevich:
Replaced with proper solution https://github.com/zowe/zen/pull/179

PR #180, opened 2024-06-12
Major Refactor

timgerstel:
1. Make sure yaml and schema are ALWAYS set
2. Change the description of smpe on the "Installation Type" stage as per @struga0258 's suggestion
5. Wrap all global variables in components in useState() to reduce memory cost *Edit: this is still here but it was actually useless lol
6. Fix scrolling on most stages
7. Fix Initialization stage status not updating when all substeps are complete
8. Add popup messages for upload yaml button on Launch Config/Networking stages
9. Make editor use entire Zowe schema
10. Use more Constants.tsx
11. Add ability to retrieve example-zowe.yaml and JSON schemas when using SMPE mode
12. Remove duplicate and unused vars
13. Fix cursor jumping to end of port input on networking stage
14. "Skip" on Unpax stage now fetches example-zowe.yaml and schemas
15. Integrate @skurnevich's method of downloading example-zowe.yaml and schemas instead of parsing cat commands🐈
16. Fix upload not working
17. Fix improper merging of example-zowe.yaml and current config thanks to @skurnevich deep merge method
18. Remove unused imports
19. Reduce duplication
20. Remove unused or unnecessary function args
21. Fix schema errors
22. Fix difference between smpe install dir and download/upload. Download/upload will no longer unpax to /runtime
23. Fix makeDir infinite loop on planning stage with invalid dirs
24. Fix yaml encoding so that zwe start can be run
25. Make editor wider
26. Fix certificates stage resetting zowe.setup.certificate values when changing zowe.verifyCertificates 
27. Increase editor width and height
28. Probably more i forgot

## Type of change
- [ ] Refactor the code 

## PR Checklist


skurnevich:
Does not work for me:

Local pax file uploading fails
Screenshot 2024-06-13 at 14 34 53


While downloading pax fails on schema parsing
```
14:43:42.429 › error setting schema from pax: SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
```

For the last one i have created the [PR 179](https://github.com/zowe/zen/pull/179) to replace cat command output parsing with files downloading. 
You can merge it first, or incorporate changes into this PR. The factual change is smaller than it looks, you just need to replace readExampleYamlAndSchema function and remove parseExampleYamlFromPax blocks as downloaded yaml and json files are ready to use

src/actions/InstallationHandler.ts
```diff
@@ -140,64 +272,17 @@ class Installation {
...
-        const yamlFromPax = parseExampleYamlFromPax(`${zoweRuntimePath}/example-zowe.yaml`);
+        const yamlFromPax = this.parseExampleYamlFromPax(`${zoweRuntimePath}/example-zowe.yaml`, readPaxYamlAndSchema.details.yaml);
         if(yamlFromPax){
           try {
             yamlObj = parse(yamlFromPax);
             if (currentConfig) {
               // console.log("current config:", JSON.stringify(currentConfig));
               // console.log("yamlObj: ", JSON.stringify(yamlObj));
-              yamlObj = {...currentConfig, ...yamlObj}
+              yamlObj = yamlObj = Object.assign({}, yamlObj, currentConfig);
```
skurnevich: Looks like a typo

src/actions/InstallationHandler.ts
```diff
@@ -140,64 +272,17 @@ class Installation {
...
-            if (yamlObj.zOSMF.host === undefined && installationArgs.zosmfHost) {
-              yamlObj.zOSMF.host = installationArgs.zosmfHost;
-            }
-            if (yamlObj.zOSMF.port === undefined && installationArgs.zosmfPort) {
-              yamlObj.zOSMF.port = installationArgs.zosmfPort;
-            }
-            if (yamlObj.zOSMF.applId === undefined && installationArgs.zosmfApplId) {
-              yamlObj.zOSMF.applId = installationArgs.zosmfApplId;
-            }
+            this.mergeYamlAndInstallationArgs(yamlObj, installationArgs);
```
skurnevich: Is there a reason why we do this merge with installationArgs values one by one, if it should have been done a line above with yamlObj = Object.assign({}, yamlObj, currentConfig); 
Is there a case when currentConfig will not exist?

src/actions/InstallationHandler.ts
```diff
@@ -50,12 +51,143 @@ class Installation {
...
+    }
+    if (installationArgs.zosmfPort) {
+      yamlObj.zOSMF.port = installationArgs.zosmfPort;
+    }
+    if (installationArgs.zosmfApplId) {
+      yamlObj.zOSMF.applId = installationArgs.zosmfApplId;
+    }
+  }
+
+  public async smpeGetExampleYamlAndSchemas (
```
skurnevich: What is different about smpe installation schemas to have this code separately?

src/renderer/components/stages/Certificates.tsx
```diff
@@ -26,28 +26,24 @@ import { createTheme } from '@mui/material/styles';
...
-  const theme = createTheme();
+  const [theme] = useState(createTheme());
 
-  const stageLabel = 'Initialization';
-  const subStageLabel = 'Certificates';
-
-  const STAGE_ID = getStageDetails(stageLabel).id;
-  const SUB_STAGES = !!getStageDetails(stageLabel).subStages;
-  const SUB_STAGE_ID = SUB_STAGES ? getSubStageDetails(STAGE_ID, subStageLabel).id : 0;
+  const [STAGE_ID] = useState(getStageDetails(INIT_STAGE_LABEL).id);
```
skurnevich, line 35: Question on this, we should use useState for variables which we want our component to be re-rendered on change.  Why use it for constants that aren't change in scope of this component?

src/renderer/components/stages/Networking.tsx
```diff
@@ -18,636 +18,545 @@ import Ajv from "ajv";
...
-            "items": {
-              "type": ["string"]
-            }
-          },
-          "externalPort": {
-            "type": "integer",
-            "minimum": 0,
-            "maximum": 65535,
-            "description": "Port number of how you access Zowe APIML Gateway from your local computer."
+const schema: any = {
```
skurnevich, line 29: What about this schema, it is zowe-yaml-schema.json right? if we need if for fallback then probably we can move it out to constants, or have fallback schemas just shipped as files with zen?

skurnevich (dismissed):

src/actions/InstallationHandler.ts
timgerstel: At one point, during the very first run of Zen current config did not exist. I haven't confirmed this behavior in a while but in my head i dont think it should ever not exist currently.

src/actions/InstallationHandler.ts
timgerstel: Very good point, i changed it to reuse this function in the downloadUnpax function to reduce duplication

src/renderer/components/stages/Certificates.tsx
timgerstel: > Why use it for constants that aren't change in scope of this component?

If it's not wrapped in useState, react will reinitialize the variable even when the component doesnt rerender I believe. If you put a naked console.log in the global scope of a React component, it will keep getting called over and over. This leak could get costly on memory so I play it safe and wrap everything in useState.

I could also be mistaken about the re-initialization part

timgerstel:
@skurnevich 

Thank you so much for your review.

>For the last one i have created the https://github.com/zowe/zen/pull/179 to replace cat command output parsing with files downloading.
You can merge it first, or incorporate changes into this PR.

I implemented this code and I love it so much more than doing string manipulations which were unreliable due to the job output not always having the same format for some reason. I implemented it into this PR.

src/renderer/components/stages/Networking.tsx
timgerstel: This schema is actually different than zowe-yaml-schema.json, it hardcodes references from the server-common.json schema because AJV was having trouble resolving references. It also has a few fixes for regexes (zoweDataset, zoweDatasetMember) that AJV did not like

src/renderer/components/stages/Certificates.tsx
timgerstel: I think I was mistaken about these variables needing to be in useState

src/actions/InstallationHandler.ts
```diff
@@ -44,18 +46,127 @@ class Installation {
...
+      const zoweRuntimePath = installationArgs.installationDir;
+      let readPaxYamlAndSchema = await this.readExampleYamlAndSchema(connectionArgs, zoweRuntimePath);
+      let parsedSchema = false, parsedYaml = false;
+      if(readPaxYamlAndSchema.details.yaml){
+        const yamlFromPax = readPaxYamlAndSchema.details.yaml;
+        if(yamlFromPax){
+          try {
+            yamlObj = parse(yamlFromPax);
+            if (currentConfig) {
+              yamlObj = Object.assign({}, yamlObj, currentConfig);
```
skurnevich: This bug comes from here
`18:45:09.025 › error parsing example-zowe.yaml: TypeError: Cannot set properties of undefined (setting 'name')`
Object.assign replaces complete zowe section so some default properties disappear, we should use some deep merge algorithm, we can use some external lib, but i would recommend to avoid adding more dependencies and would just use some recursive func like this:
```
function deepMerge(base, extension) {
  if (isObject(base) && isObject(extension)) {
    for (const key in extension) {
      if (isObject(extension[key])) {
        if (!base[key]) base[key] = {};
        deepMerge(base[key], extension[key]);
      } else {
        Object.assign(base, {[key]: extension[key]});
      }
    }
  }
  return base;
}
```

Also it worth to review other places in the code where we merge several configs...

timgerstel: Added this, another great suggestion. Really appreciate your review on this massive PR!

skurnevich:
Hi @timgerstel is this PR ready for review/merge or would you like to add more changes?

DivergentEuropeans:
Created PR: https://github.com/zowe/zen/pull/194

src/main/index.ts
```diff
@@ -195,23 +194,28 @@ const createWindow = (): void => {
...
+
+  ipcMain.handle('download-unpax', async (event, connectionArgs, installationArgs, version) => {
+    const res = await installActions.downloadUnpax(connectionArgs, installationArgs, version);
     return res;
   });
 
-  ipcMain.handle('download-unpax', async (event, connectionArgs, installationArgs, version, zoweConfig) => {
-    const res = await installActions.downloadUnpax(connectionArgs, installationArgs, version, zoweConfig);
+  ipcMain.handle('get-yaml-schema', async (event, connectionArgs, installationArgs) => {
+    const res = await installActions.smpeGetExampleYamlAndSchemas(connectionArgs, installationArgs);
```
DivergentEuropeans, line 208: shouldn't this be called "getExampleYamlAndSchemas"

I only see this used in 2 places, 1 in Stepper which is conditional to not SMPE and 2 in 
Unpax step which is also not SMPE

src/renderer/components/Home.tsx
```diff
@@ -58,12 +57,24 @@ const cards: Array<ICard> = [
...
 const makeCard = (card: ICard) => {
+  const dispatch = useAppDispatch();
   const {id, name, description, link, media} = card;
   return (  
-    <Link key={`link-${id}`} to={link}>
-      <Box sx={{ width: '40vw', height: '40vh'}}>
-        <Card id={`card-${id}`} square={true} >
+    <Link key={`link-${id}`} to={link} >
+      <Box sx={{ width: '40vw', height: '40vh'}} onClick={(e) => {
+        const flattenedData = flatten(lastActiveState);
```
DivergentEuropeans, line 73: `lastActiveState `is already a flat object with no nested properties, at a quick glance, don't see it set anywhere. Is this essential?

src/renderer/components/Home.tsx
```diff
@@ -109,6 +116,42 @@ const Home = () => {
     eventDispatcher.on('saveAndCloseEvent', () => setShowWizard(false));
 
 
+
+    //Home is the first screen the user will always see 100% of the time. Therefore, we will call the loading of the configs, schemas, and installation args here and set them to the redux memory states
```
DivergentEuropeans, line 120: Makes sense 👍

DivergentEuropeans: Home.tsx change is v satisfying

src/renderer/components/common/EditorDialog.tsx
```diff
@@ -82,23 +79,18 @@ const EditorDialog = ({contentType, isEditorVisible, toggleEditorVisibility, onC
...
 
-    const ajv = new Ajv2019()
-    ajv.addKeyword("$anchor");
-    ajv.addMetaSchema(draft7MetaSchema)
-    const validate = ajv.compile(schema);
-
     // To validate the javascript object against the schema
-    const isValid = validate(jsonData);
-    setIsSchemaValid(isValid);
+    setIsSchemaValid(!schemaValidate.errors);
```
DivergentEuropeans, line 89: N i c e

src/renderer/components/stages/Certificates.tsx
DivergentEuropeans: I haven't tested it, but I think @timgerstel might be right here. Not that you're wrong @skurnevich , I think both things are true: using const with useState introduces a small overhead. So it would be like:
For constants and simple values that do not involve expensive computation, use them directly without useState.
For values that involve expensive computation or need to persist across renders, use useState.

So something like this might be max/minning:
```
const Certificates = () => {
  const theme = useState(createTheme())[0]; // Expensive
  const stageLabel = 'Initialization';
  const STAGE_ID = useState(getStageDetails(INIT_STAGE_LABEL).id)[0]; // Potentially expensive?
};
```

src/renderer/components/stages/Certificates.tsx
```diff
@@ -63,50 +57,10 @@ const Certificates = () => {
...
-  }
-
-  if(certificateSchema) {
-    validate = ajv.compile(certificateSchema);
-  }
-
-  if(verifyCertsSchema) {
-    validateVerifyCertSchema = ajv.compile(verifyCertsSchema);
-  }
+  const [validate] = useState(() => ajv.getSchema("https://zowe.org/schemas/v2/server-base") || ajv.compile(setupSchema))
```
DivergentEuropeans, line 60: V nitpick but all of these `https://zowe.org/schemas/v2/server-base` should be Constants.BASE_SCHEMA.$id

src/renderer/components/stages/Certificates.tsx
```diff
@@ -205,13 +159,28 @@ const Certificates = () => {
     setInitClicked(true);
     updateProgress(false);
     event.preventDefault();
-    window.electron.ipcRenderer.initCertsButtonOnClick(connectionArgs, installationArgs, (await window.electron.ipcRenderer.getConfig()).details ?? yaml).then((res: IResponse) => {
-      updateProgress(res.status);
+    window.electron.ipcRenderer.initCertsButtonOnClick(connectionArgs, installationArgs).then((res: IResponse) => {
       clearInterval(timer);
-    }).catch(() => {
+      updateProgress(res.status);
+      if(!res.status){
```
DivergentEuropeans, line 165: This may be unfinished. Not that it's relevant to this PR, it's already a large and good one, but in other steps to correctly determine if we need toggleEditor, we need checks for script failure status, res.status is unreliable. Something like,

```if (res?.details && res.details.jobOutput.indexOf(JCL_UNIX_SCRIPT_OK) == -1) {```

Not sure if you had a chance to test all those res.status cases. I would have res.status = 0, but the command failed. TODO for a later PR could be fine

File: src/renderer/components/stages/InitApfAuth.tsx [hunk 13]
```diff
@@ -236,19 +196,16 @@ const InitApfAuth = () => {
         //alertEmitter.emit('showAlert', err.toString(), 'error');
         stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = true;
         stages[STAGE_ID].isSkipped = true;
-        if (typeof err === "string") {
-          console.warn('zwe init apfauth failed', err);
-        } else {
-          console.warn('zwe init apfauth failed', err?.toString()); // toString() throws run-time error on undefined or null
-        }
+        window.electron.ipcRenderer.setStandardOutput(`zwe init apfauth failed:  ${typeof err === "string" ? err : err.toString()}`).then((res: any) => {
```
DivergentEuropeans, line 199: If this was tested, then you want to add:
`alertEmitter.emit('showAlert', 'Please view Job Output for more details', 'error');`

src/renderer/components/stages/Networking.tsx
DivergentEuropeans: So if BASE_SCHEMA is AJV "fixed", can't we remove the schemas in Networking + Launch Config and use BASE_SCHEMA? Or lmk I might be missing something

src/renderer/components/stages/Networking.tsx
```diff
@@ -787,7 +684,75 @@ const Networking = () => {
...
+                        return null;
+                      })
+                      }
+                    </div>
+                  }
+                }
+              }
+               return null;
+            })}
+          </div>
```
DivergentEuropeans, line 755: This whole chunk is difficult to rea dfor me, but it looks mostly like a "code move" so it seems alright

src/renderer/components/stages/Security.tsx
```diff
@@ -234,19 +189,17 @@ const Security = () => {
       securityProceedActions(false);
       stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = true;
       stages[STAGE_ID].isSkipped = true;
-      if (typeof err === "string") {
-        console.warn('zwe init security failed', err);
-      } else {
-        console.warn('zwe init security failed', err?.toString()); // toString() throws run-time error on undefined or null
-      }
+      window.electron.ipcRenderer.setStandardOutput(`zwe init security failed:  ${typeof err === "string" ? err : err.toString()}`).then((res: any) => {
```
DivergentEuropeans, line 192: If this was tested, then you want to add:
alertEmitter.emit('showAlert', 'Please view Job Output for more details', 'error');

src/renderer/components/stages/installation/Installation.tsx
```diff
@@ -271,12 +255,12 @@ const Installation = () => {
...
       setShowProgress(true);
       dispatch(setLoading(false)); /* change skipDownload ?? false --> true to skip upload/download steps for quicker development */
-      window.electron.ipcRenderer.installButtonOnClick(connectionArgs, installationArgs, version, (await window.electron.ipcRenderer.getConfig()).details ?? yaml).then((res: IResponse) => {
+      window.electron.ipcRenderer.installButtonOnClick(connectionArgs, installationArgs).then((res: IResponse) => {
         // Some parts of Zen pass the response as a string directly into the object
         if (res.status == false && typeof res.details == "string") {
           res.details = { 3: res.details };
         }
-        if (res?.details && res.details[3] && res.details[3].indexOf(JCL_UNIX_SCRIPT_OK) == -1) { // This check means we got an error during zwe install
+        if (!res.status && res?.details && res.details[3] && res.details[3].indexOf(JCL_UNIX_SCRIPT_OK) == -1) { // This check means we got an error during zwe install
```
DivergentEuropeans, line 266: I don't think this is a good addition. I remember cases where res.status = 0, yet the command failed. That was the whole point of using 'JCL_UNIX_SCRIPT_OK' as a better source of truth over status

src/renderer/components/stages/Unpax.tsx
```diff
@@ -90,6 +89,29 @@ const Unpax = () => {
...
   }
 
+  const fetchExampleYaml = (event: any) => {
+    event.preventDefault();
+    setShowProgress(true);
+    dispatch(setDownloadUnpaxStatus(false));
+    setDownloadUnpaxProgress(downloadUnpaxStatus);
+    dispatch(setNextStepEnabled(false));
+    window.electron.ipcRenderer.fetchExampleYamlBtnOnClick(connectionArgs, installationArgs).then((res: IResponse) => {
+      if(!res.status){ //errors during runInstallation()
```
DivergentEuropeans, line 99: Same point here about unreliability of res.status

src/renderer/components/common/EditorDialog.tsx
DivergentEuropeans: that all that adjv stuff worked

DivergentEuropeans (approved):

src/renderer/components/stages/Certificates.tsx
skurnevich: No worries, @DivergentEuropeans, that was more like a question from my side. 
So you say that in this case:
`const [STAGE_ID] = useState(getStageDetails(INIT_STAGE_LABEL).id)`
`getStageDetails` function will be triggered less times than in this case: 
`const STAGE_ID = getStageDetails(INIT_STAGE_LABEL).id`
Right? Or it will bring some other performance changes? 

I am asking because i've never seen such usage before. 
For caching really expensive operations i get used to use "useMemo"

src/renderer/components/Home.tsx
timgerstel: yeah youre right but that's pretty minor i dont that one time call when the user clicks "new installation" would cause a noticeable performance impact

src/renderer/components/stages/Certificates.tsx
timgerstel: they could all be replaced yes

src/renderer/components/stages/InitApfAuth.tsx
timgerstel: That is redundant because the editor automatically pops up we don't need a popup to tell the user to look at the output that they are already looking at

src/renderer/components/stages/Security.tsx
timgerstel: >So if BASE_SCHEMA is AJV "fixed", can't we remove the schemas in Networking + Launch Config and use BASE_SCHEMA? Or lmk I might be missing something

No because the networking stage had some caveats that required its own schema which is modified from the one packaged with Zowe, maybe it can now be amended ill have to look into it

src/main/index.ts
timgerstel: sure

PR #193, opened 2024-06-18
Regression Fix + Refine Status Updates + Resolve Edge Cases + Code Cleanup

sakshibobade21:
## Proposed changes

This PR addresses Issue: 
This PR resolves a regression issue where the status for the VSAM stage was being updated without the stage being completed. It also addresses edge cases where the status was not updated as expected when stages were skipped or completed. The implementation now ensures that the status of stages and substages is properly stored and initialized accordingly. Additionally, this PR includes some code cleanup

## Type of change
- [ ] Refactor the code 

## PR Checklist


timgerstel (approved):

src/renderer/components/common/Stepper.tsx
```diff
@@ -176,7 +176,7 @@ export default function HorizontalLinearStepper({stages, initialization}:{stages
 
   const getStepIcon = (error: any, stageId: number, isSubStep?: boolean, subStepId?: number) => {
     
-    if (!error || (isSubStep && getProgress(stages[stageId].subStages[subStepId].statusKey)) || (!isSubStep && getProgress(stages[stageId].statusKey))) {
+    if ((isSubStep && getProgress(stages[stageId].subStages[subStepId].statusKey)) || (!isSubStep && getProgress(stages[stageId].statusKey))) {
```
timgerstel: Wait why dont we want to check for error

sakshibobade21: We no longer need the "error" prop. We show the warning icon depending on the skip status of the stage.

src/renderer/components/stages/progress/StageProgressStatus.ts
```diff
@@ -274,19 +291,58 @@ export const mapAndGetSkipStatus = (subStageId: number): boolean => {
   return skipStatusArray[subStageId];
 }
 
-export const setSubStageSkipStatus = (key: keyof SkipState, newValue: boolean): void => {
+export const mapAndSetStepSkipStatus = (stageId: number, value: boolean): void => {
+  setStageSkipStatus(stepSkipKeysArray[stageId-1], value);
```
timgerstel, line 286: this needs a check to make sure stageId isn't an out of bounds index

src/renderer/components/stages/progress/StageProgressStatus.ts
```diff
@@ -274,19 +291,58 @@ export const mapAndGetSkipStatus = (subStageId: number): boolean => {
...
+  const skipStatus = getStageSkipStatus();
+  const skipStatusArray = [
+    skipStatus.planning,
+    skipStatus.installationType,
+    skipStatus.unpax,
+    skipStatus.initialization,
+    skipStatus.reviewInstallation
+  ]
+
+  return skipStatusArray[subStageId-1];
```
timgerstel: check if array index out of bounds

src/renderer/components/common/Stepper.tsx
timgerstel: Should probably remove the arg then if its not needed

sakshibobade21: Yes it is removed.

src/renderer/components/stages/progress/StageProgressStatus.ts
sakshibobade21: Done. Thanks

src/renderer/components/stages/progress/StageProgressStatus.ts
sakshibobade21: Done. Thanks

src/renderer/components/stages/Security.tsx
```diff
@@ -187,8 +186,6 @@ const Security = () => {
       updateProgress(false);
       clearInterval(timer);
       securityProceedActions(false);
-      stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = true;
-      stages[STAGE_ID].isSkipped = true;
```
DivergentEuropeans, line 192: Can you explain these changes a bit more? So why are we removing some of these skipped statuses when things fail? If they are green, and are re-run and they fail, status should go from green to skipped no? (well ideally "error" status but it's nbd)

src/renderer/components/stages/Networking.tsx
```diff
@@ -569,8 +569,6 @@ const Networking = () => {
 
     dispatch(setNextStepEnabled(getProgress('networkingStatus')));
     dispatch(setInitializationStatus(isInitComplete()));
-    stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = false;
-    stages[STAGE_ID].isSkipped = isInitializationSkipped;
```
DivergentEuropeans, line 573: My comment applies here too

src/renderer/components/stages/InitApfAuth.tsx
```diff
@@ -194,8 +192,6 @@ const InitApfAuth = () => {
         updateProgress(false);
         // TODO: Test this
         //alertEmitter.emit('showAlert', err.toString(), 'error');
-        stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = true;
-        stages[STAGE_ID].isSkipped = true;
```
DivergentEuropeans, line 199: My comment applies here too

src/renderer/components/stages/installation/Installation.tsx
```diff
@@ -269,21 +270,14 @@ const Installation = () => {
...
-          stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = !res.status;
-          setNextStepEnabled(true);
           clearInterval(timer);
         }
       }).catch((err: any) => {
         clearInterval(timer);
         updateProgress(false);
-        installProceedActions(false);
-        stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = true;
-        stages[STAGE_ID].isSkipped = true;
```
DivergentEuropeans, line 296: Here too, etc. etc.

DivergentEuropeans: Also did you mean to remove 'installProceedActions' too?

src/renderer/components/stages/Certificates.tsx
```diff
@@ -129,7 +129,7 @@ const Certificates = () => {
   const setStageSkipStatus = (status: boolean) => {
     stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = status;
     stages[STAGE_ID].isSkipped = status;
-    mapAndSetSkipStatus(SUB_STAGE_ID, status);
+    mapAndSetSubStepSkipStatus(SUB_STAGE_ID, status);
```
DivergentEuropeans: If we have `stages` which is state, and we can set both status for the main step and sub step, so looks like we need a `mapAndSetSubStepSkipStatus` for setting local storage, ok sure

but I think
 a) this method should be renamed to like "setAndSaveSubStepSkipStatus"/"setAndStoreSubStepSkipStatus" and "setAndSaveStepSkipStatus"/"setAndStoreStepSkipStatus" because it's doing a storage operation
 b) if these methods already do the setting, why do we need `stages[STAGE_ID].xxxx` stuff at all, shouldn't it be removed
 c) doesn't need to be this PR, but why is `StageProgressStatus.ts` not using a Store?

src/renderer/components/stages/Security.tsx
sakshibobade21: If things fail, we call updateProgress(false). Update progress internally calls setStageSkipStatus(true) which sets the status for the stages.

sakshibobade21: This changes were made to cleanup the code and remove redundant lines of code

src/renderer/components/stages/Networking.tsx
sakshibobade21: If things fail, we call updateProgress(false). Update progress internally calls setStageSkipStatus(true) which sets the status for the stages.

This changes were made to cleanup the code and remove redundant lines of code

src/renderer/components/stages/installation/Installation.tsx
sakshibobade21: Removed `installProceedActions()` because updateProgress() is calling it internally.

Again this change was made to cleanup the code and remove redundant lines of code

src/renderer/components/stages/Certificates.tsx
sakshibobade21: The function mapAndSetSubStepSkipStatus() is used to update the local storage. stages[STAGE_ID].xxxx modifies the state only for the current session. When the user resumes, we utilize mapAndGetSubStepSkipStatus() to update stage[stage_id].skip.

Throughout the session, stages[STAGE_ID].skip is used to retrieve and update the status. While mapAndGetSubStepSkipStatus() is useful for fetching status, doing so from local storage every time could impact performance negatively.

sakshibobade21: I agree with you; there is indeed scope for code cleanup and improvement here. We can simplify by calling mapAndSetSubStepSkipStatus(). Within this function, we can set stages[STAGE_ID].skip and update the local storage.

Could we handle the code cleanup in a separate PR?

sakshibobade21: We also have` mapAndSetStepSkipStatus()` for stages.
I can rename these to `setAndSaveSubStepSkipStatus` but we already have `setSubStageSkipStatus`.
`mapAndSetSubStepSkipStatus` calls `setSubStageSkipStatus` by passing the `stageid` and the `status`.

DivergentEuropeans: > While mapAndGetSubStepSkipStatus() is useful for fetching status, doing so from local storage every time could impact performance negatively.

This is mostly irrelevant because we (should) only update the status after button clicks & operations. If this is negatively affecting performance, that means we are calling it during each render operation which points to an issue elsewhere

> I agree with you; there is indeed scope for code cleanup and improvement here. We can simplify by calling mapAndSetSubStepSkipStatus().

Yes, let's move forward with that and I would prefer to have it done in this PR because you already started some of that duplication removal. So just continue that for everywhere, or do it in a separate PR, but don't begin it & then leave it unfinished (unless you write a detailed TODO)

Rename sounds good 👍 setSubStageSkipStatus and setAndStoreSubStepSkipStatus / setAndSaveSubStepSkipStatus both sound good and make sense, for what they are titled for

DivergentEuropeans: If your setAndSave methods work properly, then the refactor is just a text search and replace so I would prefer it done in the same PR

src/renderer/components/stages/Networking.tsx
DivergentEuropeans: Where do we call it near this code? Not called in this file at least

src/renderer/components/stages/Certificates.tsx
DivergentEuropeans: Then you would want to check for performance hits, by seeing if the useEffect( )'s or whatever are called after significant necessary updates versus each re-render. It could be out of scope to FIX all the performance hits, but what it means is:
some methods will have
```
stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = status;
    stages[STAGE_ID].isSkipped = status;
```


bla bla bla right, some methods will have
```
setAndStoreSubStepSkipStatus(SUB_STAGE_ID, status);
```

but no methods should have BOTH

src/renderer/components/stages/Networking.tsx
sakshibobade21: I removed those lines of code because every time the networking component loads, it incorrectly sets the `skip` status to false. If the networking stage is skipped and the user revisits it, the status should be skipped

sakshibobade21: Same applies for the launch config stage

src/renderer/components/stages/Certificates.tsx
sakshibobade21: We have

 ```
const setStageSkipStatus = (status: boolean) => {
    stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = status;
    stages[STAGE_ID].isSkipped = status;
    mapAndSetSubStepSkipStatus(SUB_STAGE_ID, status);
  }
```

`.isSkipped` is just updating the stages array.
And `mapAndSetSubStepSkipStatus` is updating the local storage.

We can move below code
```
stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = status;
stages[STAGE_ID].isSkipped = status;
```

inside `mapAndSetSubStepSkipStatus`. But I think it is better we keep them separated.

Comment (sakshibobade21) [reply src/renderer/components/stages/Certificates.tsx]: I will try to address the performance improvements in the same PR.

Comment (sakshibobade21) [reply src/renderer/components/stages/Certificates.tsx]: We have
```
const setStageSkipStatus = (status: boolean) => {
   stages[STAGE_ID].subStages[SUB_STAGE_ID].isSkipped = status;
   stages[STAGE_ID].isSkipped = status;
   mapAndSetSubStepSkipStatus(SUB_STAGE_ID, status);
 }
```
`.isSkipped` is just updating the stages array and not the local storage.
And `mapAndSetSubStepSkipStatus` is updating the local storage.

We need to update both.

**Comment by sakshibobade21** [comment]
@DivergentEuropeans I addressed all possible code cleanups, both related and unrelated to the stage statuses. I also tried to optimized performance by minimizing calls to functions that interact with local storage, such that these functions are called only once when the component unmounts. Additionally, I renamed the functions we discussed earlier.

File: src/renderer/components/common/Stepper.tsx [hunk 9]
```diff
@@ -240,7 +229,19 @@ export default function HorizontalLinearStepper({stages, initialization}:{stages
...
+        const completion = getProgress(stages[activeStep].subStages[activeSubStep].statusKey);
+        return !completion;
+      } else {
+        const completion = getProgress(stages[activeStep].statusKey);
+        return !completion;
+      }
+    } else {
+      return true;
+    }
+  }
```
DivergentEuropeans: I am a bit confused by this change. So originally, we set `isNextSStepEnabled` based on useAppSelector(selectNextStepEnabled) which comes from the `setNextStepEnabled` state which is a thing where steps have logic to signal if next step is enabled, based on if a certain action has been performed. Ok, sure. 

But now, we are moving away from this to --> check for a `getProgress` which comes from local storage, where a status is being set based on logic within steps upon status completion.

a) So if case 1's state is no longer needed, why don't we remove it? 

b) Why is it called "getProgress"/"setProgress" instead of "get/setStatus" ? 

c) StageProgressStatus.ts in my opinion is a bit long and over-engineered and difficult to read. I think I approved a smaller, earlier version of it many months ago but this was a big mistake. Why are there useless keys there like, 
`let installationTypeKey = 'installation_type';` when this could be 
`INSTALLATION_TYPE_STAGE_LABEL` from Utils.tsx (which was previously just a const in a home/main page somewhere)

You don't need to do anything with c) in this PR

DivergentEuropeans: And btw for c) I'm not talking about localstorage vs redux. That's fine that we used local storage, but the file is still unnecessarily difficult to read

src/renderer/components/stages/Certificates.tsx
```diff
@@ -144,10 +151,15 @@ const Certificates = () => {
...
-    dispatch(setInitializationStatus(isInitComplete()));
-    dispatch(setCertificateStatus(status));
     setCertificateInitializationProgress(getCertificateInitState());
+    dispatchActions(status);
+  }
+
+  const dispatchActions = (status: boolean) => {
+    dispatch(setCertificateStatus(status));
+    dispatch(setInitializationStatus(isInitializationStageComplete()));
+    dispatch(setNextStepEnabled(status));
```
DivergentEuropeans, line 166: If we are now using "Status" as the signal that tells us if next step is enabled, then all of these `setNextStepEnabled` now become useless no?

DivergentEuropeans (dismissed):

src/renderer/components/stages/Certificates.tsx
sakshibobade21: Himani identified an issue in my pull request: the next step was not enabled during the unpax stage when the stage was completed.
While debugging, I discovered that the nextStepButtonDisabled function was being called multiple times, even when the stage status remained unchanged. This caused the data to be fetched from local storage repeatedly. Although this didn't significantly impact performance, it would be more efficient to use the setNextStepEnabled function instead.
Therefore, I updated the code to use setNextStepEnabled, along with some modifications.

src/renderer/components/common/Stepper.tsx
sakshibobade21: Himani identified an issue in my pull request: the next step was not enabled during the unpax stage when the stage was completed.
While debugging, I discovered that the nextStepButtonDisabled function was being called multiple times, even when the stage status remained unchanged. This caused the data to be fetched from local storage repeatedly. Although this didn't significantly impact performance, it would be more efficient to use the setNextStepEnabled function instead.
Therefore, I updated the code to use setNextStepEnabled, along with some modifications.

skurnevich (approved):
So in the current state we do not distinguish between failed and skipped steps?  

Also, for instance if Certificates step fails - it shows job output as error, but then we skip it and go to vsam (which should be completely optional) and it fails too - it will show error in the header, and the job log stays from the previous failure in Certificates. It looks inconsistent and misleading. 

I do not mind merging this though, as it doesn't break things and in order to continue the development process.

src/renderer/components/stages/Certificates.tsx
sakshibobade21: Addressed - https://github.com/zowe/zen/pull/193#discussion_r1668324613

sakshibobade21:
@skurnevich Thank you for reviewing the PR.

At present, in ZEN, we do not differentiate between failed and skipped steps. This could be a potential enhancement to consider for future improvements.

Regarding your comment about the certificate stage, I think we should address that in a separate PR since it originates from the staging. I will create a ticket for this.

Thanls.

skurnevich:
Hi Sakshi, that is fine with me, please resolve Lenny's changes request and we can merge this PR

sakshibobade21:
@skurnevich thanks. And the changes requested by Lenny have already been addressed.

Also, I do not have the merge rights for this repository. Could you please merge it if you are ok with it? This PR will serve as a foundation for future PRs in ZEN.

skurnevich:
Hi @sakshibobade21 , i'll merge this PR, but problem is the github still has "Changes requested" as blocker for merge there.
Screenshot 2024-08-06 at 11 11 15
As far as i can see the requests were addressed, and I can merge it bypassing the protection if @DivergentEuropeans  does not mind. 
But the proper way to resolve this is to re-request review, 
Screenshot 2024-08-06 at 11 25 24
Github sometimes acts weird with permissions, so can you please check if you have this option in the top right corner first?

sakshibobade21:
@skurnevich I re-requested the review now. Hope that works.
Thank you.

DivergentEuropeans:
Thanks, dismissed stale review @sakshibobade21 @skurnevich

PR #195, opened 2024-06-24
Added word wrap to the Editor (helps with viewing 1 line errors)

DivergentEuropeans:
This makes the errors received in places like zwe install (the `[0: { nskdansndanjsd [....] }]` ones) a lot more readable when the error is really long 1 line

We still have work to do to show errors better in UI but this helps.


DivergentEuropeans:
This was added

PR #202, opened 2024-06-30
Rest yaml and schema fetching progress

timgerstel:
## Proposed changes

- Fix fetch/re-fetch yaml + schema progress UI not resetting
- Fix app reload resetting to connection screen causing schema runtime errors
- Fix password saving to electron-store. Password is now NEVER stored by accident.
- Fix VSAM text input fields scrambling user input

## Type of change
- [ ] Refactor the code 

## PR Checklist


src/actions/ConnectionHandler.ts
```diff
@@ -36,8 +36,8 @@ export class FTPConnection extends Connection {
         response.details = e.message;
       }
     }
+    config.password = "";
```
DivergentEuropeans, line 39: I am very intrigued by how this fixes things, or it at least implies this stuff

``` 
if(k != "password"){
        const status = ConnectionStore.set(`ftp-details.${k}`, value);
        return acc + status ? '' : `\n Can't set ftp-details.${k}, check the store schema`;
      }
 ```
      
 isn't great. I feel like we're covering up something but youre not making it worse

DivergentEuropeans (approved):

src/actions/ConnectionHandler.ts
timgerstel: Nothing is being covered up, the password is needed for when the connection is made to z/OS and then should be thrown away before storing the rest of the FTP connection details (host, username, etc).

DivergentEuropeans: "Nothing is being covered up", completely untrue - check again @timgerstel 

saveConnectionData( ) shows `config` object gets parsed. Except even if `password` key exists, no value is stored in a Store (is intended implementation). Yet your bugfix removes it further upstream which is fine. But the original code should have prevented any kind of bug, it doesn't (the snippet I showed), and should be removed or at least a comment that says "//TODO: Not working as intended"

DivergentEuropeans: > Fix password saving to electron-store

timgerstel: Its not a bug, the password was being saved because the whole connection store was being set to config, which is the connection details directly from the Connection stage UI

> Yet your bugfix removes it further upstream which is fine
This is literally as early as it can be removed because the call literally right before is connectFTPServer which obviously needs the password

timgerstel: I dont even know what you mean by "cover up" could you elaborate

PR #204, opened 2024-07-05
Bugfix - Missing Password

sakshibobade21:
## Proposed changes

This PR addresses Issue:
Adds a dialog for entering a password when the resume button is clicked to fixe the bug where the password was missing from the PASS command.

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/src/renderer/components/common/passwordDialog.tsx
+++ b/src/renderer/components/common/passwordDialog.tsx
@@ -34,7 +34,6 @@ const PasswordDialog = ({ onPasswordSubmit }:{ onPasswordSubmit: any }) => {
   };
 
   const onPasswordUpdate = (password: string) => {
-    setIsLoading(true);
     dispatch(setPassword(password));
   }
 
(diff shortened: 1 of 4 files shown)


src/renderer/components/common/passwordDialog.tsx
```diff
@@ -0,0 +1,112 @@
+/*
```
timgerstel, line 1: Should probably reuse the connection component instead of adding a new one

timgerstel:
Im confused on what this PR does. When you click the resume feature, we already show the connection screen if there is no valid session. So this PR now adds...a second login screen?

sakshibobade21:
@timgerstel 

https://github.com/zowe/zen/issues/184
https://jira.rocketsoftware.com/browse/MVD-6437

This dialog is specifically for resuming within the same session. The details you mentioned above pertain to resuming in a different session.

As we discussed before, we aim to avoid saving the password, even within the redux state, correct @1000TurquoisePogs  ?

Therefore, when the user clicks 'Save and Close,' we will remove the password. For resuming within the same session, only the password is required, not the other fields from the connection tab. Hence, the purpose of this dialog

src/renderer/components/common/passwordDialog.tsx
DivergentEuropeans: @timgerstel I think you are fundamentally misunderstanding what this PR is about

DivergentEuropeans: Ah, I see

>  When you click the resume feature, we already show the connection screen if there is no valid session. 

This part was not working for same session. Sakshi fixed this + added more minimal UI

PR #208, opened 2024-07-09
Created workaround for New Install not resetting during same session

DivergentEuropeans:
after clicking Save & Close


1000TurquoisePogs (approved):
Fine for now - thank you for adding code comment on what needs to be fixed later.

PR #218, opened 2024-07-11
Regression fixes for 1st Release

sakshibobade21:
## Proposed changes

This PR fixes the regressions:
1) The "Resume" option appears on the home page of ZEN even when no previous installation exists.
2) Each time the YAML is updated in the editor, the cursor automatically moves to the beginning of the file.

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/src/renderer/components/Home.tsx
+++ b/src/renderer/components/Home.tsx
@@ -203,7 +203,7 @@ const Home = () => {
         setIsNewInstallation(true);
       } else {
         const data: ActiveState = unflatten(JSON.parse(lastInstallation));
-        setIsNewInstallation(false);
+        setIsNewInstallation(!(data && data.lastActiveDate));
       }
 
 
--- a/src/renderer/components/common/EditorDialog.tsx
+++ b/src/renderer/components/common/EditorDialog.tsx
@@ -46,9 +46,6 @@ const EditorDialog = ({contentType, isEditorVisible, toggleEditorVisibility, onC
     /* TODO: Should use an array for the Store to house separate outputs (Security vs Certificates for example) */
     if(isEditorVisible) { 
        if(contentType == 'yaml') {
-        if (setupYaml?.installationArgs) {
-          delete setupYaml.installationArgs;
-        }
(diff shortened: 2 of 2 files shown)


src/renderer/components/Home.tsx
```diff
@@ -203,7 +203,7 @@ const Home = () => {
         setIsNewInstallation(true);
       } else {
         const data: ActiveState = unflatten(JSON.parse(lastInstallation));
-        setIsNewInstallation(false);
+        setIsNewInstallation(!(data && data.lastActiveDate));
```
timgerstel, line 206: I changed this line because quite frequently I would have a previously saved installation not show the resume button due to lastActiveDate being missing. Please fix the root cause instead (lastActiveDate not always being set properly.

timgerstel (changes requested):

src/renderer/components/Home.tsx
sakshibobade21: Never encountered an issue where - "resume button not showing sometimes when a previous installation does exist"
Confirmed this with Himani.
The previous date is always saved in the activeStage object, except for the home page where this behavior is intentional.

src/renderer/components/common/EditorDialog.tsx
```diff
@@ -46,6 +46,9 @@ const EditorDialog = ({contentType, isEditorVisible, toggleEditorVisibility, onC
     /* TODO: Should use an array for the Store to house separate outputs (Security vs Certificates for example) */
     if(isEditorVisible) { 
        if(contentType == 'yaml') {
+        if (setupYaml?.installationArgs) {
```
timgerstel: This second delete is unnecessary, remove this and leave the other one

sakshibobade21: How is this unnecessary?

timgerstel: Because its redundant you only need the other delete

timgerstel (approved):

PR #231, opened 2024-07-19
ZEN - Multiple bugfixes

sakshibobade21:
## Proposed changes

This PR addresses several issues:
1) States of the stages not resetting correctly during a new reinstallation.
2) Unnecessary restart occurring when clicking 'Save and Close'.
3) Missing password dialog when clicking 'Save and Close'.
4) Job statement missing during new installation.
5) Incorrect status display in the connection tab.
6) Adding the warning dialog for new Installation

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/src/renderer/components/Dialogs/WarningDialog.tsx
+++ b/src/renderer/components/Dialogs/WarningDialog.tsx
@@ -1,0 +1,43 @@
+/*
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ */
+import { Box, Button, Dialog, DialogActions, DialogContent, DialogTitle, FormControl, TextField } from '@mui/material';
+import { useState } from 'react';
+
+const WarningDialog = ({onWarningDialogSubmit}: {onWarningDialogSubmit: any}) => {
+
+  const [isDialogVisible, setIsDialogVisible] = useState(true);
+
+  const handleClose = () => {
+    setIsDialogVisible(false);
+    onWarningDialogSubmit(false);
+  }
+
+  const handleSubmit = () => {
+    setIsDialogVisible(false);
+    onWarningDialogSubmit(true);
+  }
+
+  return (
+    <div>
+      <Dialog open={isDialogVisible} maxWidth={'xs'}>
+        <DialogTitle>Warning!</DialogTitle>
+        <DialogContent> Hey There </DialogContent>
+        <DialogActions>
+          <div>
+            <Button onClick={handleClose}>Submit</Button>
+            <Button onClick={handleSubmit}>Close</Button>
+          </div>
+        </DialogActions>
+      </Dialog>
+    </div>
+  )
+}
+
+export default WarningDialog;
--- a/src/renderer/components/Dialogs/WarningDialog.tsx
+++ b/src/renderer/components/Dialogs/WarningDialog.tsx
@@ -31,8 +31,8 @@ const WarningDialog = ({onWarningDialogSubmit}: {onWarningDialogSubmit: any}) =>
         <DialogContent> Hey There </DialogContent>
         <DialogActions>
           <div>
-            <Button onClick={handleClose}>Submit</Button>
-            <Button onClick={handleSubmit}>Close</Button>
+            <Button onClick={handleSubmit}>Submit</Button>
+            <Button onClick={handleClose}>Close</Button>
           </div>
         </DialogActions>
       </Dialog>
--- a/src/renderer/components/Dialogs/WarningDialog.tsx
+++ b/src/renderer/components/Dialogs/WarningDialog.tsx
@@ -28,10 +28,10 @@ const WarningDialog = ({onWarningDialogSubmit}: {onWarningDialogSubmit: any}) =>
     <div>
       <Dialog open={isDialogVisible} maxWidth={'xs'}>
         <DialogTitle>Warning!</DialogTitle>
-        <DialogContent> Hey There </DialogContent>
+        <DialogContent style={{ fontSize: '13px', fontFamily: 'Arial, sans-serif', color: 'black' }}> Starting a new installation will erase the previously stored installation data from the wizard. Do you wish to proceed? </DialogContent>
         <DialogActions>
           <div>
-            <Button onClick={handleSubmit}>Submit</Button>
+            <Button onClick={handleSubmit}>Proceed</Button>
             <Button onClick={handleClose}>Close</Button>
           </div>
         </DialogActions>
--- a/src/renderer/components/Home.tsx
+++ b/src/renderer/components/Home.tsx
@@ -28,7 +28,7 @@ import Wizard from './configuration-wizard/Wizard'
 import { ActiveState } from '../../types/stateInterfaces';
 import { getInstallationArguments, getPreviousInstallation } from './stages/progress/StageProgressStatus';
 import { DEF_NO_OUTPUT, FALLBACK_SCHEMA, FALLBACK_YAML } from './common/Utils';
-import { selectInstallationArgs, setInstallationArgs, installationSlice, setIsNewerInstallation } from './stages/installation/installationSlice';
+import { selectInstallationArgs, setInstallationArgs, installationSlice, setIsNewInstallation, selectIsNewInstallation } from './stages/installation/installationSlice';
 import PasswordDialog from './common/passwordDialog';
 
 // REVIEW: Get rid of routing
@@ -76,13 +76,13 @@ const Home = () => {
 
   const { activeStepIndex, isSubStep, activeSubStepIndex, lastActiveDate } = getPreviousInstallation();
 
-  const [isNewInstallation, setIsNewInstallation] = useState(false);
-
   const stages: any = [];
   const defaultTooltip: string = "Resume";
   const resumeTooltip = connectionStatus ? defaultTooltip : `Validate Credentials & ${defaultTooltip}`;
   const [showPasswordDialog, setShowPasswordDialog] = useState(false);
   const [isResume, setIsResume] = useState(useAppSelector(selectResumeProgress));
+
+  const isNewInstallation = useAppSelector(selectIsNewInstallation);
 
   const makeCard = (card: ICard) => {
     const {id, name, description, link, media} = card;
@@ -90,8 +90,7 @@ const Home = () => {
     const handleClick = () => {
       let newInstallationArgs = installationSlice.getInitialState().installationArgs;
       if (id === "install") {
-        dispatch(setIsNewerInstallation(true));
-        setIsNewInstallation(true);
+        dispatch(setIsNewInstallation(true));
         dispatch(setConnectionStatus(false));
         newInstallationArgs = {...newInstallationArgs, dryRunMode: false};
       } else if (id === "dry run") {
@@ -201,10 +200,10 @@ const Home = () => {
       if (!lastInstallation) {
         const flattenedData = flatten(lastActiveState);
         localStorage.setItem(prevInstallationKey, JSON.stringify(flattenedData));
-        setIsNewInstallation(true);
+        dispatch(setIsNewInstallation(true));
       } else {
         const data: ActiveState = unflatten(JSON.parse(lastInstallation));
-        setIsNewInstallation(!(data && data.lastActiveDate));
+        dispatch(setIsNewInstallation(!(data && data.lastActiveDate)));
       }
 
 
@@ -217,7 +216,7 @@ const Home = () => {
   const resumeProgress = () => {
     setShowWizard(true);
     dispatch(setResumeProgress(true));
-    dispatch(setIsNewerInstallation(false));
+    dispatch(setIsNewInstallation(false));
 
     if(connectionStatus) {
       setShowPasswordDialog(true);
@@ -234,7 +233,7 @@ const Home = () => {
       {!showWizard && <div className="home-container" style={{ display: 'flex', flexDirection: 'column' }}>
 
         <div style={{ position: 'absolute', left: '-9999px' }}>
-          <HorizontalLinearStepper stages={stages} />
+          {stages.length > 0 && <HorizontalLinearStepper stages={stages} />}
         </div>
 
         {!connectionStatus && <div style={{marginBottom: '20px'}}></div>}
--- a/src/renderer/components/Home.tsx
+++ b/src/renderer/components/Home.tsx
@@ -28,7 +28,7 @@ import Wizard from './configuration-wizard/Wizard'
 import { ActiveState } from '../../types/stateInterfaces';
 import { getInstallationArguments, getPreviousInstallation } from './stages/progress/StageProgressStatus';
 import { DEF_NO_OUTPUT, FALLBACK_SCHEMA, FALLBACK_YAML } from './common/Utils';
-import { selectInstallationArgs, setInstallationArgs, installationSlice } from './stages/installation/installationSlice';
+import { selectInstallationArgs, setInstallationArgs, installationSlice, setIsNewerInstallation } from './stages/installation/installationSlice';
 import PasswordDialog from './common/passwordDialog';
 
 // REVIEW: Get rid of routing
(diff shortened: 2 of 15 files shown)


src/renderer/components/Home.tsx
```diff
@@ -268,8 +272,8 @@ const Home = () => {
     {showWizard &&
       <>
         {showPasswordDialog && <PasswordDialog onPasswordSubmit={confirmConnection}></PasswordDialog>}
-        {(showPasswordDialog && updatedConnection) && <Wizard initialization={false}/>}
-        {!showPasswordDialog && <Wizard initialization={false}/>}
+        {(showPasswordDialog && updatedConnection) && <Wizard initialization={isNewInstallation}/>}
+        {!showPasswordDialog && <Wizard initialization={isNewInstallation}/>}
```
skurnevich, line 286: These two lines look like over-complication, what is the use of updatedConnection? Why we can't use showPasswordDialog only? 
Also there is inconsistency, if you resume progress after click Save & Close it will show the dialog, but if you close the app and click resume after re-opening it will start with Connection tab. 
And line 73 has unused values `const [showLoginDialog, setShowLogin] = useState(false);` that are probably intended for the same purpose?

src/renderer/components/Home.tsx
```diff
@@ -88,10 +88,13 @@ const Home = () => {
 
   const makeCard = (card: ICard) => {
     const {id, name, description, link, media} = card;
-  
+
     const handleClick = () => {
       let newInstallationArgs = installationSlice.getInitialState().installationArgs;
       if (id === "install") {
+        dispatch(setIsNewerInstallation(true));
+        setIsNewInstallation(true);
```
skurnevich: Having setIsNewInstallation and setIsNewerInstallation looks like an algorithmic issue, what is the reason to have both?

src/renderer/components/common/Stepper.tsx
```diff
@@ -17,7 +17,7 @@ import StepLabel from '@mui/material/StepLabel';
 import Button from '@mui/material/Button';
 import Typography from '@mui/material/Typography';
 import { Link } from 'react-router-dom';
-import { selectConnectionStatus } from '../stages/progress/progressSlice';
+import { selectConnectionStatus, setConnectionStatus } from '../stages/progress/progressSlice';
```
skurnevich: setConnectionStatus does not seem to be used here, and should not be set from this component

src/renderer/components/common/Stepper.tsx
```diff
@@ -29,25 +29,31 @@ import Warning from '@mui/icons-material/Warning';
 import CheckCircle from '@mui/icons-material/CheckCircle';
 import { TYPE_YAML, TYPE_OUTPUT, TYPE_JCL, INIT_STAGE_LABEL, REVIEW_INSTALL_STAGE_LABEL, UNPAX_STAGE_LABEL } from '../common/Utils';
 import { getProgress, getCompleteProgress, mapAndSetSkipStatus, mapAndGetSkipStatus } from '../stages/progress/StageProgressStatus';
-
+import {initProgressStatus} from '../stages/progress/progressConst';
```
skurnevich: So is this a set of defaults or replacement for the progress storage? Shouldn't defaults be part of the storage anyway?

src/renderer/components/common/Stepper.tsx
```diff
@@ -29,25 +29,31 @@ import Warning from '@mui/icons-material/Warning';
...
 import { selectConnectionArgs, setPassword } from '../stages/connection/connectionSlice';
-import { selectInstallationArgs } from '../stages/installation/installationSlice';
+import { selectInstallationArgs, selectIsNewInstallation } from '../stages/installation/installationSlice';
 // TODO: define props, stages, stage interfaces
 // TODO: One rule in the store to enable/disable button
 
 export default function HorizontalLinearStepper({stages, initialization}:{stages: any, initialization?:boolean}) {
 
-  const connectionStatus = useSelector(selectConnectionStatus);
+  const connectionStatus = useState(useAppSelector(selectConnectionStatus));
```
skurnevich: Why the useSelector even appeared there over the useAppSelector? If it was by mistake i still can see it in the line 58 `    useSelector(selectConnectionStatus),`
And again, any benefits of using useState here?

src/renderer/components/common/Stepper.tsx
```diff
@@ -29,25 +29,31 @@ import Warning from '@mui/icons-material/Warning';
...
   const REVIEW_STAGE_ID = getStageDetails(REVIEW_INSTALL_STAGE_LABEL).id;
 
-  const completeProgress = getCompleteProgress();
+  const isNewInstallation = useAppSelector(selectIsNewInstallation);
+
+  let completeProgress = {...initProgressStatus};
 
+  if(!isNewInstallation) {
+    completeProgress = getCompleteProgress();
+  }
```
skurnevich: I suggest to get rid of initProgressStatus, but even if not the ternary operator can be used here instead of 'completeProgress' re-assigning

src/renderer/components/stages/progress/StageProgressStatus.ts
```diff
@@ -166,10 +68,48 @@ const setKeys = (id: string) => {
...
+export const resetProgress = (host: string, user: string,) => {
+  if(host && user) {
+    const id = `${host}_${user}`;
+    setKeys(id);
+  }
+
+  const keysArray = [progressStateKey, activeStateKey, planningStateKey, installationTypeKey, downloadUnpaxKey, datasetInstallationKey, apfAuthKey, securityKey, stcsKey, certificateKey, vsamKey, planningValidationDetailsKey, prevInstallationKey, skipStateKey, installationArgsKey]; 
+  keysArray.forEach(key => {
+    if(localStorage.getItem(key) !== null) {
+      localStorage.removeItem(key);
```
skurnevich, line 100: So all of above is happening just to reset the storage in case user click on New Zowe Installation when they have some Saved Installation? Can you share more details on how this works with the keys and ids?

skurnevich (changes requested):
Also when i use this PR "Upload Zowe PAX" never stops spinning, even when step is done and "Continue to Components Installation" button is unlocked, not sure why that happens but it wasn't like that when i tried PR 193 
Screenshot 2024-08-02 at 15 03 22

sakshibobade21:
> Also when i use this PR "Upload Zowe PAX" never stops spinning, even when step is done and "Continue to Components Installation" button is unlocked, not sure why that happens but it wasn't like that when i tried PR 193 Screenshot 2024-08-02 at 15 03 22

@skurnevich
The issue originates from v2.x/staging, and my PR specifically addresses this problem - https://github.com/zowe/zen/pull/193

For reference, please see the attached screenshot:

image (41)

src/renderer/components/Home.tsx
sakshibobade21: The following behavior is intentional:

1) If we click "Save & Close" and then resume in the same session, a dialog prompts for the password. Since we do not store the password in the app, we only ask for it. 
Additionally, because it is the same session, the user does not need to see the entire connection tab—just the password is required.

2) If we restart the app (different session) and resume, the connection tab with all attributes is displayed before resuming.

This behavior originates from staging.

sakshibobade21: Great catch @skurnevich regarding the `updatedConnection`. This issue originates from staging and has been addressed in this PR now.

Thanks.

src/renderer/components/Home.tsx
sakshibobade21: I added `setIsNewerInstallation` as a Redux state to be used globally across other stages. 
The `isNewInstallation` is from v2.x/staging and it is not necessary. 
I just updated the PR to rely solely on the Redux state.

Thanks!

src/renderer/components/common/Stepper.tsx
sakshibobade21: Removed it.

src/renderer/components/common/Stepper.tsx
sakshibobade21: Removed `initProgressStatus`

src/renderer/components/common/Stepper.tsx
sakshibobade21: Removed `initProgressStatus`. And added: `const completeProgress = getCompleteProgress();`
`getCompleteProgress()` takes care of it.

src/renderer/components/common/Stepper.tsx
sakshibobade21: Ideally, it should be `useAppSelector`. This originated from staging, and I'm not sure how or when `useSelector` was added. I've fixed it in this PR. Additionally, `useState` is not necessary for `connectionStatus` here since it is not responsible for updating any state. So removed it.

src/renderer/components/Home.tsx
sakshibobade21: Additionally, I have removed unused states and variables in the Home component that originated from staging.

src/renderer/components/common/Stepper.tsx
sakshibobade21: Additionally, I have removed unused states, functions and variables in the Stepper component that originated from staging.

src/renderer/components/stages/progress/StageProgressStatus.ts
sakshibobade21: The code is designed to reset various progress-related states and storage entries when a user initiates a new Zowe installation, even if they have saved installation data. The `setProgressObjects` function initializes all the necessary state objects to their default values, ensuring that the application starts fresh. The `resetProgress` function clears the relevant keys from local storage to remove any saved data associated with the previous installation. This process ensures that the application doesn't carry over any outdated or conflicting data from a previous setup. By using unique IDs based on the `host` and `user`, the code manages and isolates storage and state for different user sessions.

skurnevich: Ok, i still believe it is not correct, your idea is to split different sessions in the storage so user can continue with multiple instances installation is great and we should have it some eventually. But `host` and `user` combination is not enough to define a unique installation, for that we need to use unique session/instance id. Second, as of now it doesn't even work for me, when i save my session and then ask different user to start new one they still can see some values from my installation, and after that i can't return to my first installation as there is just one card for saved installation. 
So in the current state i propose to keep multiple saved installations as a feature to be done in future, and just when user clicks on New Installation to show a warning popup that their previous progress will be deleted and after that actually delete all the contents of configuration storage

src/renderer/components/Home.tsx
skurnevich: Ok thank you, sorry for me asking on things that are not part of this PR, but even if it is already in staging we still need to fix things that are not in a good shape. 
So i do believe that we need to unify this different behavior for resuming the saved installation, because even if the app is reopened we sill know the host/port and user for that installation, so we just need to ask for password and validate the entire connection. 
I can't see any reason to ask user for host or to change it, moreover if user gets to connection tab and changes the host for saved installation it will corrupt the entire installation as part of progress is done on a different machine. 
So ok let's not have it as part of this PR as it does not belongs here, but i'll create an issue for fixing that.

skurnevich (changes requested):

src/renderer/components/Home.tsx
sakshibobade21: Sure @skurnevich Thank you for the detailed review on my PRs.

sakshibobade21: We previously discussed these two different approaches for the resume. However, we can revisit and refine them further.
Thanks for creating the issue.

src/renderer/components/stages/progress/StageProgressStatus.ts
sakshibobade21: Yes, using just the host and username cannot fully distinguish between different installations, so this approach needs reworking. However, due to time constraints, we planned to proceed with this for the first release.
For future enhancements, we plan to support multiple saved installations. 

In the meantime, I will implement a warning popup. Thanks again.

sakshibobade21: @skurnevich Added the warning dialog for the new installation

image

src/renderer/components/Dialogs/WarningDialog.tsx
```diff
@@ -0,0 +1,43 @@
...
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ */
+import { Box, Button, Dialog, DialogActions, DialogContent, DialogTitle, FormControl, TextField } from '@mui/material';
+import { useState } from 'react';
+
+const WarningDialog = ({onWarningDialogSubmit}: {onWarningDialogSubmit: any}) => {
```
skurnevich, line 13: This seems to be quite hardcoded again. 
First it is better to have the WarningDialog.tsx in "common", no need to have separate structure for that. 
Second we should have it as simple as possible, taking 3 properties "text", "onSubmit", "onCancel" and no logic inside, even visibility should be controlled in the parent component. In that case it could be reused in future in any other place. We actually can go even further and add 4th property to control dialog type, like error/warning/info etc.

src/renderer/components/Home.tsx
```diff
@@ -22,25 +22,22 @@ import { Tooltip } from '@mui/material';
...
+import { ActiveState, InstallationArgs } from '../../types/stateInterfaces';
 import { getInstallationArguments, getPreviousInstallation } from './stages/progress/StageProgressStatus';
 import { DEF_NO_OUTPUT, FALLBACK_SCHEMA, FALLBACK_YAML } from './common/Utils';
-import { selectInstallationArgs, setInstallationArgs, installationSlice } from './stages/installation/installationSlice';
+import { selectInstallationArgs, setInstallationArgs, installationSlice, setIsNewInstallation, selectIsNewInstallation } from './stages/installation/installationSlice';
 import PasswordDialog from './common/passwordDialog';
+import WarningDialog from './Dialogs/WarningDialog';
+import HomeCardComponent from './HomeCardComponent';
+import { ICard } from '../../types/interfaces';
+import { ROUTES } from '../../Routes/RouteConstant';
```
skurnevich, line 36: Similar with Routes, you have moved it out to a separate file structure Routes/RouteConstant.tsx which looks like an overkill to me, especially when there are just two links and routing is not actually needed, it is just a legacy leftover that should be cleaned up later. And definitely we do not need Routes outside of renderer. 
It is fine to move constants that need to be shared between modules out to some distinct file, if there are not many, but in this case you can use something like generic constants.ts in the “renderer” or “src” folder. 
Also cards constant still uses text "/wizard" on lines 46 and 53

src/renderer/components/Home.tsx
```diff
@@ -216,33 +171,93 @@ const Home = () => {
...
+      dispatch(setResumeProgress(false));
+    }
+    handleNewInstallation(newInstallationArgs);
+  };
+
+  const confirmNewInstallation = (status: boolean) => {
+    dispatch(setIsNewInstallation(status))
+    setNewInstallationClick(false);
+    setPreviousInstallation(!status);
+
```
skurnevich, line 222: Well, it does not clear previous information for me, it starts with a connection screen again, but all the data still there. I was going to recommend you to just drop the stores here, like every store has the deleteAll() function for that purpose, but when i tried that i realize it also has no effect until you close the app. 
I believe we have some issues with the way we use storage, we duplicate it in memory and so just complicate things, creating multiple sources of data

src/renderer/components/stages/progress/StageProgressStatus.ts
```diff
@@ -10,126 +10,23 @@
 
 import { flatten, unflatten } from 'flat';
 import { ProgressState, PlanningState, InstallationType, ActiveState, DatasetInstallationState, InitSubStepsState, CertInitSubStepsState, PlanningValidationDetails, SkipState, InstallationArgs, DownloadUnpaxState} from '../../../../types/stateInterfaces';
+import { initProgressStatus, initInstallationTypeStatus, initDownloadUnpaxStatus, initActiveStatus, initPlanningStageStatus, initDatasetInstallationStatus, initApfAuthStatus, initSecurityInitStatus, initStcsInitStatus, initCertificateInitStatus, initVsamInitStatus, initPlanningValidationDetailsStatus, initStepSkipStatus, initInstallationArgsStatus } from './progressConst';
```
skurnevich, line 13: Why we even have StageProgressStatus.ts, why don't we use storage/ProgressStore.ts for that?

skurnevich (changes requested):

PR #232, opened 2024-07-26
Bugfix - Security Tab

sakshibobade21:
## Proposed changes

This PR addresses Issue: 
When the security tab is launched for the first time, this PR populates the fields with the default values

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/src/renderer/components/common/JsonForms.tsx
+++ b/src/renderer/components/common/JsonForms.tsx
@@ -107,28 +107,15 @@ const conditionalSchema = (schema: any, formData: any, prop: any): boolean=> {
 
 export default function JsonForm(props: any) {
   const {schema, onChange, formData} = props;
-  const [localFormData, setLocalFormData] = useState(formData || {});
 
-  // const isFormDataEmpty = formData === null || formData === undefined || Object.keys(formData).length < 1;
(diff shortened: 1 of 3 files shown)


src/renderer/components/common/JsonForms.tsx
```diff
@@ -108,7 +108,7 @@ const conditionalSchema = (schema: any, formData: any, prop: any): boolean=> {
 export default function JsonForm(props: any) {
   const {schema, onChange, formData} = props;
 
-  const isFormDataEmpty = formData === null || formData === undefined;
+  const isFormDataEmpty = formData === null || formData === undefined || Object.keys(formData).length < 1;
 
   useEffect(() => {
```
skurnevich: The JsonForm is a stateless component and being rendered based on props anyway so useEffect does not make any sense here to me. Can you please describe a bit more about what we are trying to cover here? Case with security when we do not have defaults in the yaml file so we take it from schema?

sakshibobade21: That's right. When we do not have values for any attributes in security, we display the defaults from the schema.

sakshibobade21: Yes, JsonForm must be stateless and should not include useEffect. This issue originates from the staging environment. I have addressed it in this PR.

skurnevich (approved):

PR #233, opened 2024-07-29
Dataset Validation On Installation Stage

sakshibobade21:
## Proposed changes

This PR addresses Issue: This PR adds the dataset validations on the installation stage

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/src/renderer/components/stages/installation/Installation.tsx
+++ b/src/renderer/components/stages/installation/Installation.tsx
@@ -29,6 +29,8 @@ import { getStageDetails, getSubStageDetails } from "../../../../services/StageD
 import { getProgress, setDatasetInstallationState, getDatasetInstallationState, getInstallationTypeStatus, mapAndSetSkipStatus, getInstallationArguments, datasetInstallationStatus, isInitComplete } from "../progress/StageProgressStatus";
 import { DatasetInstallationState } from "../../../../types/stateInterfaces";
 import eventDispatcher from '../../../../services/eventDispatcher';
+import { isDatasetValid } from '../../../../services/DatasetValidation';
+import ErrorIcon from '@mui/icons-material/Error';
 
 const Installation = () => {
 
@@ -67,7 +69,6 @@ const Installation = () => {
   const [installationType] = useState(getInstallationTypeStatus().installationType);
 
   const [validate] = useState(() => ajv.getSchema("https://zowe.org/schemas/v2/server-base") || ajv.compile(setupSchema));
-
   
   useEffect(() => {
     dispatch(setInitializationStatus(isInitComplete()));
@@ -221,6 +222,7 @@ const Installation = () => {
     stages[STAGE_ID].isSkipped = status;
     mapAndSetSkipStatus(SUB_STAGE_ID, status);
   }
+
   const updateProgress = (status: boolean) => {
     setStateUpdated(!stateUpdated);
     setStageSkipStatus(!status);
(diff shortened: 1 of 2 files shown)


src/renderer/components/stages/installation/Installation.tsx
```diff
@@ -300,6 +308,31 @@ const Installation = () => {
     updateProgress(true);
   }
   }
+
+  const validateDatasets = () => {
+    for (const key of Object.keys(setupYaml)) {
+      let value = setupYaml[key];
+
+      if (typeof value === 'object' && value !== null) {
```
skurnevich: I guess this part is hardcoded to the current state of zowe.yaml, assuming that there is just one value in the parmlibMembers object - {zis: ZWESIP00}. 
If we add one more value there later it would not be validated, so instead we should recursively go to validateDatasets with that new object or flatten data

src/renderer/components/stages/installation/Installation.tsx
```diff
@@ -300,6 +308,31 @@ const Installation = () => {
...
+      }
+
+      const isValid = isDatasetValid(value);
+      if (!isValid) {
+        setIsFormValid(false);
+        setFormError(`The dataset '${key.toUpperCase()}' is invalid. Please verify the dataset name and try again.`);
+        return false;
+      }
+    }
+    if(Object.keys(setupYaml).length < 8) {
```
skurnevich: Same here, now we have 8 values there but it can easily change, moreover for instance zowe.setup.dataset.authLoadlib is actually optional. 
This better to be controlled by schema, but i can't see it for zowe.setup.dataset, so i guess for now we can remove hardcoded numbers and just validate it in the UI by not allowing empty values?

src/renderer/components/stages/installation/Installation.tsx
```diff
@@ -357,8 +390,12 @@ const Installation = () => {
         </Typography>
 
         <Box sx={{ width: '60vw' }} onBlur={async () => dispatch(setYaml((await window.electron.ipcRenderer.getConfig()).details ?? yaml))}>
-          {!isFormValid && formError && <div style={{color: 'red', fontSize: 'small', marginBottom: '20px'}}>{formError}</div>}
           <JsonForm schema={setupSchema} onChange={handleFormChange} formData={setupYaml}/>
+          {!isFormValid && formError &&
+            <div style={{ display: 'flex', marginBottom: '10px' }}>
+              <ErrorIcon color="error" sx={{ fontSize: 20 }}/>
```
skurnevich: I thought we have some unified errors popup in the header? why we have different approach here?

skurnevich (changes requested):

src/renderer/components/stages/installation/Installation.tsx
sakshibobade21: Updates this. Thank you.

src/renderer/components/stages/installation/Installation.tsx
sakshibobade21: Addressed these changes. Thank you.

src/renderer/components/stages/installation/Installation.tsx
sakshibobade21: I initially considered displaying the form errors right below the form after clicking "Install MVS Datasets." However, let's stick to using the unified error messages in the header instead.

Update this. Thank you.

skurnevich (approved):
LGTM, however, in future i would prefer it to be validated via schema like https://github.com/zowe/zowe-install-packaging/blob/v2.x/staging/schemas/server-common.json#L39 
But we need to update schema for that

PR #237, opened 2024-08-01
[WIP] ZEN Bugfix for the Schema Updates

sakshibobade21:
## Proposed changes

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/src/actions/InstallationHandler.ts
+++ b/src/actions/InstallationHandler.ts
@@ -159,12 +159,21 @@ class Installation {
             if(yamlSchema && serverCommon){
               yamlSchema.additionalProperties = true;
               yamlSchema.properties.zowe.properties.setup.properties.dataset.properties.parmlibMembers.properties.zis = zoweDatasetMemberRegexFixed;
-              yamlSchema.properties.zowe.properties.setup.properties.certificate.properties.pkcs12.properties.directory = serverCommon.$defs.path;
+
+              if(yamlSchema.properties?.zowe?.properties?.setup?.properties?.certificate?.properties?.pkcs12?.properties?.directory) {
+                yamlSchema.properties.zowe.properties.setup.properties.certificate.properties.pkcs12.properties.directory = serverCommon.$defs.path;
+              } else if (yamlSchema.properties?.zowe?.properties?.setup?.properties?.certificate?.oneOf) {
+                const certificateSchemaArray = yamlSchema.properties.zowe.properties.setup.properties.certificate.oneOf;
+                const filteredArray = certificateSchemaArray.filter((element:any) => element.required && element.required.includes('pkcs12'));
+                if(filteredArray.length === 1) {
+                  filteredArray[0].properties.pkcs12.properties.directory = serverCommon.$defs.path;
+                }
+              }
+
               if(yamlSchema.$defs?.networkSettings?.properties?.server?.properties?.listenAddresses?.items){
                 delete yamlSchema.$defs?.networkSettings?.properties?.server?.properties?.listenAddresses?.items?.ref;
                 yamlSchema.$defs.networkSettings.properties.server.properties.listenAddresses.items = serverCommon.$defs.ipv4
               }
-              // console.log('Setting schema from runtime dir:', JSON.stringify(yamlSchema));
               ConfigurationStore.setSchema(yamlSchema);
               parsedSchema = true;
               ProgressStore.set('downloadUnpax.getSchemas', true);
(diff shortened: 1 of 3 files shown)


src/actions/InstallationHandler.ts
```diff
@@ -159,12 +159,21 @@ class Installation {
...
               yamlSchema.properties.zowe.properties.setup.properties.dataset.properties.parmlibMembers.properties.zis = zoweDatasetMemberRegexFixed;
-              yamlSchema.properties.zowe.properties.setup.properties.certificate.properties.pkcs12.properties.directory = serverCommon.$defs.path;
+
+              if(yamlSchema.properties?.zowe?.properties?.setup?.properties?.certificate?.properties?.pkcs12?.properties?.directory) {
+                yamlSchema.properties.zowe.properties.setup.properties.certificate.properties.pkcs12.properties.directory = serverCommon.$defs.path;
+              } else if (yamlSchema.properties?.zowe?.properties?.setup?.properties?.certificate?.oneOf) {
+                const certificateSchemaArray = yamlSchema.properties.zowe.properties.setup.properties.certificate.oneOf;
+                const filteredArray = certificateSchemaArray.filter((element:any) => element.required && element.required.includes('pkcs12'));
+                if(filteredArray.length === 1) {
+                  filteredArray[0].properties.pkcs12.properties.directory = serverCommon.$defs.path;
```
timgerstel, line 169: This looks like it will correctly set the directory definition, but the original yamlSchema object needs to be updated too since filteredArray is a separate variable.

src/actions/InstallationHandler.ts
```diff
@@ -159,12 +159,21 @@ class Installation {
             if(yamlSchema && serverCommon){
               yamlSchema.additionalProperties = true;
               yamlSchema.properties.zowe.properties.setup.properties.dataset.properties.parmlibMembers.properties.zis = zoweDatasetMemberRegexFixed;
-              yamlSchema.properties.zowe.properties.setup.properties.certificate.properties.pkcs12.properties.directory = serverCommon.$defs.path;
+
+              if(yamlSchema.properties?.zowe?.properties?.setup?.properties?.certificate?.properties?.pkcs12?.properties?.directory) {
+                yamlSchema.properties.zowe.properties.setup.properties.certificate.properties.pkcs12.properties.directory = serverCommon.$defs.path;
```
1000TurquoisePogs, line 164: why is the schema being modified?
this will only lead to disappointment, such as the break we see in v2.17
there will be more in the future.
don't edit the schema, and it wont break.

1000TurquoisePogs (changes requested):

src/actions/InstallationHandler.ts
sakshibobade21: Updating the filteredArray will also update yamlSchema because filteredArray contains references to objects within the yamlSchema structure.
What you are saying is valid for the primitive values.

src/actions/InstallationHandler.ts
sakshibobade21: The code is not modifying the schema. It is only setting the reference for `pkcs12.directory`

src/actions/InstallationHandler.ts
timgerstel: Just tested that out you are correct👍

src/actions/InstallationHandler.ts
1000TurquoisePogs: When you consider that this is doing a $ref resolve https://github.com/zowe/zowe-install-packaging/blob/1f9ad776820f62672d7b3e17d64ef8b9b0569332/schemas/zowe-yaml-schema.json#L176C62-L176C70 then I think a better way to prevent it from breaking in the future is to recurse over the entire `yamlSchema` object looking for keys with name `$ref` and then doing a swap of their values depending on what the value is.
The value `...#` is the `$anchor` property value of another object.

src/actions/InstallationHandler.ts
skurnevich: Even while this is working, it is still a poor approach, it does mutate an object and does it in an unobvious way, in the best case scenario - code just becomes less readable, worst case it may later lead to some bugs that are really hard to spot. 
I would recommend to use some setter function, but i guess this part is not applicable anymore as this code should be gone in final version where we map all the $refs for the entire yamlSchema object in advance, is that correct?

sakshibobade21: @skurnevich here is the PR to address this issue - https://github.com/zowe/zen/pull/240
Thanks.

PR #241, opened 2024-08-09
 Update JSON Forms

sakshibobade21:
## Proposed changes

This PR addresses Issue: 
1) Updates json forms to work with the schemas with oneOf attribute
2) Adds the radio buttons to choose from the available schemas
3) Conditionally showing the "Vsam Dataset" field only if "name" does not exists on the vsam object
4) The "name" field in the vsam object updates the attribute components.caching-service.storage.vsam.name
5) Updated the dataset pattern to allow 38 characters ate max. Basically I replaced the previous dataset pattern  to the one over here: https://github.com/zowe/zowe-install-packaging/blob/v2.x/staging/schemas/server-common.json#L31 so that we have consistency
6) Code Organization and Cleanup

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/src/actions/InstallationHandler.ts
+++ b/src/actions/InstallationHandler.ts
@@ -121,7 +121,6 @@ class Installation {
       let parsedSchema = false, parsedYaml = false;
       if(readPaxYamlAndSchema.details.yaml){
         const yamlFromPax = readPaxYamlAndSchema.details.yaml;
-        let yamlSchema;
         if(yamlFromPax){
           try {
             yamlObj = parse(yamlFromPax);
@@ -146,7 +145,7 @@ class Installation {
         //No reason not to always set schema to latest if user is re-running installation
         if(readPaxYamlAndSchema.details.schemas.yamlSchema){
           try {
-            yamlSchema = JSON.parse(readPaxYamlAndSchema.details.schemas.yamlSchema);
+            let yamlSchema = JSON.parse(readPaxYamlAndSchema.details.schemas.yamlSchema);
             updateSchemaReferences(readPaxYamlAndSchema.details.schemas, yamlSchema);
             if(yamlSchema){
               ConfigurationStore.setSchema(yamlSchema);
@@ -165,7 +164,7 @@ class Installation {
           ConfigurationStore.setSchema(FALLBACK_SCHEMA);
           return {status: false, details: {message: 'no schemas found from pax'}}
         }
-        return {status: parsedSchema && parsedYaml, details: {message: "Successfully retrieved example-zowe.yaml and schemas", mergedYaml: yamlObj, yamlSchema: yamlSchema}}
+        return {status: parsedSchema && parsedYaml, details: {message: "Successfully retrieved example-zowe.yaml and schemas", mergedYaml: yamlObj}}
       }
     } catch (e) {
       ConfigurationStore.setSchema(FALLBACK_SCHEMA);
@@ -247,16 +246,14 @@ class Installation {
         }
       }
       let yamlObj = {};
-      let yamlSchema = {};
       await this.getExampleYamlAndSchemas(connectionArgs, installationArgs).then((res: IResponse) => {
         if(res.status){
           if(res.details.mergedYaml != undefined){
             yamlObj = res.details.mergedYaml;
-            yamlSchema = res.details.yamlSchema;
           }
         }
       })
-      return {status: download.status && uploadYaml.status && upload.status && unpax.status, details: {message: 'Zowe unpax successful.', mergedYaml: yamlObj, yamlSchema: yamlSchema}};
+      return {status: download.status && uploadYaml.status && upload.status && unpax.status, details: {message: 'Zowe unpax successful.', mergedYaml: yamlObj}};
     } catch (error) {
       return {status: false, details: error.message};
     }
--- a/src/actions/InstallationHandler.ts
+++ b/src/actions/InstallationHandler.ts
@@ -21,15 +21,6 @@ import { InstallationArgs } from '../types/stateInterfaces';
 import { FALLBACK_SCHEMA, deepMerge } from '../renderer/components/common/Utils';
 import { updateSchemaReferences } from '../services/ResolveRef';
 
-//AJV did not like the regex in our current schema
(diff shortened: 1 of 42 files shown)


src/actions/InstallationHandler.ts
```diff
@@ -154,17 +155,10 @@ class Installation {
         //No reason not to always set schema to latest if user is re-running installation
         if(readPaxYamlAndSchema.details.yamlSchema && readPaxYamlAndSchema.details.serverCommon){
           try {
-            let yamlSchema = JSON.parse(readPaxYamlAndSchema.details.yamlSchema);
+            yamlSchema = JSON.parse(readPaxYamlAndSchema.details.yamlSchema);
             const serverCommon = JSON.parse(readPaxYamlAndSchema.details.serverCommon);
+            updateSchemaReferences(readPaxYamlAndSchema.details, yamlSchema);
```
skurnevich: Why are we duplicating changes made in https://github.com/zowe/zen/pull/240 here?

src/actions/InstallationHandler.ts
```diff
@@ -181,7 +175,7 @@ class Installation {
           ConfigurationStore.setSchema(FALLBACK_SCHEMA);
           return {status: false, details: {message: 'no schemas found from pax'}}
         }
-        return {status: parsedSchema && parsedYaml, details: {message: "Successfully retrieved example-zowe.yaml and schemas", mergedYaml: yamlObj}}
+        return {status: parsedSchema && parsedYaml, details: {message: "Successfully retrieved example-zowe.yaml and schemas", mergedYaml: yamlObj, yamlSchema: yamlSchema}}
```
skurnevich: Schema is already saved to the ConfigurationStore, lines 162, 169 and 175. 
Then we take yamlSchema that will exist only in successful case (as you only set it in the try block on line 157 and so it will be empty in other cases instead of FALLBACK_SCHEMA)
And then you return it to the UI (Unpax.tsx:180) where use for rendering and override the ConfigurationStore scheme if it exists, which makes not much sense to me, or am i missing something?

src/renderer/components/common/JsonForms.tsx
```diff
@@ -13,38 +13,117 @@ import { JsonForms } from '@jsonforms/react';
...
+const createHorizontalLayout = (elements: any[]) => ({
+  type: 'HorizontalLayout',
+  elements,
+});
 
-const makeUISchema = (schema: any, base: string, formData: any): any => {
-  if (!schema || !formData) {
-    return "";
+// To handle the "if", "else", and "then" in the schema
+const conditionalSchema = (schema: any, formData: any, prop: any): boolean=> {
```
skurnevich, line 49: What reference were you using for this definition? I am looking at https://json-schema.org/understanding-json-schema/reference/conditionals#ifthenelse and it states that `then` or `else` can be omitted, also `then` and `else` sections can have `properties` with pattern, not only `required` field

src/renderer/components/common/JsonForms.tsx
```diff
@@ -86,64 +185,63 @@ const makeUISchema = (schema: any, base: string, formData: any): any => {
...
-    <JsonForms
-      schema={schema}
-      uischema={makeUISchema(schema, '/', formData)}
-      data={formDataToUse}
-      renderers={materialRenderers}
-      cells={materialCells}
-      config={{showUnfocusedDescription: true}}
-      onChange={({ data, errors }) => { onChange(data) }}
-    />
+      {  schema.oneOf &&
```
skurnevich, line 214: General question, can't JSONForms work with "oneOf" out of the box? 
I didn't research much but from [here](https://jsonforms.io/docs/renderer-sets) it seems that is should be supported using tabs?

skurnevich (changes requested):

src/actions/InstallationHandler.ts
sakshibobade21: This PR is built on the top of #240

src/actions/InstallationHandler.ts
sakshibobade21: That's a good catch. Thanks. I fixed it.

src/renderer/components/common/JsonForms.tsx
sakshibobade21: This section was initially added to handle if-else conditions in the certificate schema for previous versions of Zowe. At that time, we were only verifying whether the conditional attributes were met.

These sections can also contain properties, and while all the properties were displayed, our primary focus was on checking if the condition matched.

src/renderer/components/common/JsonForms.tsx
sakshibobade21: It could work, but we had a slightly different approach in mind:

1. We preferred radio buttons instead of tabs.
2. The JSON form data is stored in YAML, so we needed the YAML to comply with only one schema at a time. For this, we created a function called filterFormData.
3. We also required AJV to validate the data against the correct schema selected from the oneOf schemas.
4. Additionally, we don't use the JsonForms renderer as-is. We customize it as needed in the makeUISchema function. To make this work, we had to take this approach

src/renderer/components/stages/InitApfAuth.tsx
```diff
@@ -231,21 +230,7 @@ const InitApfAuth = () => {
       }, {});
     }
 
-    if(validate) {
```
skurnevich, line 234: Why did you remove this validation here? There is unused code block a few lines above, `newData` is never being used, so i wonder if it is not a mistake?

skurnevich (approved):
Ok, but please review the comment to [InitApfAuth.tsx](https://github.com/zowe/zen/pull/241#discussion_r1742169008) before merge

src/renderer/components/stages/InitApfAuth.tsx
sakshibobade21: I removed this part of the code because it is unnecessary. The APF authorization stage includes read-only fields populated during the previous "Dataset Initialization" stage, where validations for these fields have already been performed. Re-validating them here is redundant.

I also considered removing the formChangeHandler function, as it's no longer required. However, I chose not to do so in this PR to keep the focus solely on the schema updates. I plan to address it in future code optimization PRs.

PR #271, opened 2025-02-21
Bug fix for artifactory redirect

struga0258:
Artifcatory has moved to a redirect download method and current library does not automatically handle redirects. This pr detects the 3xx respond and redirects to new url. Fixing the bug where convivence build installation fails since it never downloads the Zowe pax.


timgerstel (approved):
Change looks good, should make Zen compatible with URL redirects for pax download (untested on my machine, only reviewed changes)

1000TurquoisePogs:
lets put a 10 redirect limit in.

1000TurquoisePogs (approved):

PR #272, opened 2025-02-28
Make Server Install wizard download with Zowe V3 and V2

DivergentEuropeans:
PR title self explanatory! Makes the Downloading work with V3 and V2

image

However, V3 and V2 need to be tested. I don't think both versions are compatible with wizard atm, though some people reported successes in the past. The hardcoding in our schemas and YAML implies at least some bugs with getting correct configuration translated from wizard to z/OS, but those could have been fixed by now

- also  makes z/OS Unix path runtime check better 
- use optional arg to specify what zowe vers to get example-zowe.yaml, schema, or minor (full) version string


src/actions/InstallationHandler.ts
```diff
@@ -219,8 +219,11 @@ class Installation {
       }
 
       let download, upload, unpax;
-      if(installationArgs.installationType === "download"){
-        console.log("downloading...", version);
+      if(installationArgs.installationType === "downloadV2" || installationArgs.installationType === "downloadV3"){
+        if (installationArgs.installationType === "downloadV2") {
+          version = ZOWE_V2_LATEST;
```
1000TurquoisePogs: Absolutely not.
Just parse the contents of https://zowe.jfrog.io/artifactory/libs-release-local/org/zowe/ to find the latest version of v2. You are not going to hardcode releases.

1000TurquoisePogs (dismissed):

src/renderer/components/stages/CachingService.tsx
```diff
@@ -379,6 +379,7 @@ const CachingService = () => {
             </FormControl>
           }
 
+          {/* TODO: Add support for ZOWE V3 'INFINISPAN' */}
```
DivergentEuropeans: image

src/renderer/components/stages/Planning.tsx
```diff
@@ -244,11 +245,11 @@ const Planning = () => {
           alertEmitter.emit('showAlert', details.error, 'error');
         }
       }).finally(() => {
-        // TODO: Make it more smart, add these checks to the planning validation details and verify the input fields one by one with a status icon.
-        if (localYaml?.zowe?.logDirectory.startsWith(localYaml?.zowe?.runtimeDirectory) ||
-            localYaml?.zowe?.workspaceDirectory.startsWith(localYaml?.zowe?.runtimeDirectory) ||
-            localYaml?.zowe?.extensionDirectory.startsWith(localYaml?.zowe?.runtimeDirectory)) {
-            alertEmitter.emit('showAlert', `Some instance locations (workspace, logs or extensions) are defined inside the runtime directory ${localYaml?.zowe?.runtimeDirectory}. It is not recommended as the runtime directory ment to be read-only.`, 'warning', 20000);
+        // TODO: Add these checks to the planning validation details and verify the input fields one by one with a status icon.
```
DivergentEuropeans, line 248: This is an acoustic change 🏆  I just made the checks themselves more smart, but kept the original comment

struga0258 (approved):

